PR #44, opened 2023-02-09
Minimal security threeshold

adi90x:
Hello, 

As discussed, I have add a user defined threshold for the security preset to be activated
I can put an "input" instead of fixed value not sure we can have both available ? 

Regards,

### DIFF
--- a/README-fr.md
+++ b/README-fr.md
@@ -248,7 +248,7 @@ Le premier délai (minimal_activation_delay_sec) en sec dans le délai minimum a
 Le deuxième délai (security_delay_min) est le délai maximal entre deux mesures de température avant de régler le préréglage sur ``security`` et d'éteindre le thermostat. Si le capteur de température ne donne plus de mesures de température, le thermostat et le radiateur s'éteindront après ce délai et le préréglage du thermostat sera réglé sur ``security``. Ceci est utile pour éviter une surchauffe si la batterie de votre capteur de température est trop faible.
 
 Le troisième paramétre (security_min_on_percent) est la valeur minimal de on_percent en dessous de laquelle le préréglage sécurité ne sera pas activé.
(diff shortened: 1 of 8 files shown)


jmcollin78:
Hello @adi90x 

Many thanks for your PR. I will quickly have a look to this, and I came back with eventual comments.

jmcollin78 (approved):
Fine !

jmcollin78:
I will integrate your code. The errors were there before and I am currently fixing it.

adi90x:
Thanks a lot ! 
One quick point I have just seen, it seems a bit strange the way the interface react when type in the field I have added ( it seems it didn't like input that start by 0.xxx  not sure if you have seen it ?

jmcollin78:
@adi90x 

When using your PR on existing thermostat, I've got this error:

```
  File "/workspaces/core/config/custom_components/versatile_thermostat/climate.py", line 1179, in _async_temperature_changed
    await self._async_control_heating(force=False)
  File "/workspaces/core/config/custom_components/versatile_thermostat/climate.py", line 1740, in _async_control_heating
    security: bool = await self.check_security()
  File "/workspaces/core/config/custom_components/versatile_thermostat/climate.py", line 1670, in check_security
    and self._prop_algorithm.on_percent > self._security_min_on_percent
TypeError: '>' not supported between instances of 'int' and 'NoneType'
```

We will have to intialise the _security_min_on_percent with the default value if None to be upward compatible.
I will add a quick fix.

jmcollin78:
> One quick point I have just seen, it seems a bit strange the way the interface react when type in the field I have added ( it seems it didn't like input that start by 0.xxx not sure if you have seen it ?

Yes it seems to be a bug in the interface with '.'. 0.0 seems to be not accepted.

adi90x:
> > One quick point I have just seen, it seems a bit strange the way the interface react when type in the field I have added ( it seems it didn't like input that start by 0.xxx not sure if you have seen it ?
> 
> Yes it seems to be a bug in the interface with '.'. 0.0 seems to be not accepted.

Quite strange but it seems to be the same with TPI coeff also so I guess we can close that :)

PR #86, opened 2023-06-12
Update sensor.py

Bergoglio:
Added target temperature as a sensor to be displayed in the sensors window and be used as a variable in other programs.

### DIFF
--- a/custom_components/versatile_thermostat/sensor.py
+++ b/custom_components/versatile_thermostat/sensor.py
@@ -47,6 +47,7 @@ async def async_setup_entry(
         LastTemperatureSensor(hass, unique_id, name, entry.data),
         LastExtTemperatureSensor(hass, unique_id, name, entry.data),
         TemperatureSlopeSensor(hass, unique_id, name, entry.data),
+        TargetTemperatureSensor(hass, unique_id, name, entry.data),
     ]
     if entry.data.get(CONF_DEVICE_POWER):
         entities.append(EnergySensor(hass, unique_id, name, entry.data))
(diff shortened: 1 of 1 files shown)


custom_components/versatile_thermostat/sensor.py
```diff
@@ -423,4 +424,41 @@ def native_unit_of_measurement(self) -> str | None:
...
+    @property
+    def device_class(self) -> SensorDeviceClass | None:
+        return SensorDeviceClass.TEMPERATURE
+
+    @property
+    def state_class(self) -> SensorStateClass | None:
+        return SensorStateClass.MEASUREMENT
+
+    @property
+    def native_unit_of_measurement(self) -> str | None:
```
jmcollin78, line 463: Here you should use the Home Assistant setting of the user. See climate.py :

```
@property
    def temperature_unit(self) -> str:
        """Return the unit of measurement."""
        if self._is_over_climate and self.underlying_entity(0):
            return self.underlying_entity(0).temperature_unit

        return self._unit
```


and:

`self._unit = self._hass.config.units.temperature_unit`

custom_components/versatile_thermostat/sensor.py
```diff
@@ -423,4 +424,41 @@ def native_unit_of_measurement(self) -> str | None:
     @property
     def suggested_display_precision(self) -> int | None:
         """Return the suggested number of decimal digits for display."""
-        return 2
+        return
```
jmcollin78, line 427: Why this change ? Should not be good because an int is expected in return

jmcollin78 (changes requested):
Thank you for your PR. 2 simples remarks and I will merge this PR.

jmcollin78:
No more action on this. I will close the PR. Don't hesitate to reopen it if you think it is still relevant

PR #112, opened 2023-10-01
Update climate.py

diplix:
without these parenthesis the `action` and therefore returned `hvac_action` seems to be `True` instead of i.e. `heating` (did not see that when using an underlying homekit devices climate device, but when the underlying was a generic thermostat)


jmcollin78:
Thank you for this message. I will have a look on this rapidly

jmcollin78:
You was right, thank you !

Issue #146, opened 2023-10-30
Mean Power cycle


adi90x:
Hello, 

Seems that when using 2 ( or more ) heater entity, power consumption input in configuration is multiply by ```n``` heater.
However, if both heater are not same power it leads to a wrong power for the climate ( with impact on power management )
We should either configure each 4 underlying heater with their own power , or maybe just configure total power in configuration no ? 

Good evening ,

jmcollin78:
Hello @adi90x,

Your second proposal make sense. Having the total of power since the better option. I don't want to configure each heater separately.

adi90x:
Could be usefull to have separate info, only if we decide to implement smart power management, like one big and one small heater could be more interesting to switch on maybe only small for 100% instead of 30% small/30% big. But not even sure it is interesting to do that ( as if there is several heater I guess we prefer all at 30% than one full and other switch off !)

adi90x:
Closing with #150

PR #150, opened 2023-11-01
Mean power Update

adi90x:
Hello, 
Reference Issue  #146 
Mean Power shouldn't multiply device_power by number of heater registered.
Also add a small information on that in the documentation.


jmcollin78:
Hello @adi90x ,

Thank you for your PR. I wonder why no unit tests have changed. It seems not normal if the rule changed.

adi90x:
@jmcollin78 I guess there is no unit test for that 😁

jmcollin78:
> @jmcollin78 I guess there is no unit test for that 😁

Does test_power.py pass ?
If you change something, something must break or please add a test that shows what you have done works.

adi90x:
> > @jmcollin78 I guess there is no unit test for that 😁
> 
> Does test_power.py pass ? If you change something, something must break or please add a test that shows what you have done works.
> 
> EDIT : I try with your change and all pass without error. So can you please add a test that shows you change works ? In long terms, it is essential to have a good coverage of the code. Thank you in advance. You have the test_power.py in example if you want.

I think that it pass because test is done with only one heater so there is no difference in the ```mean_cycle_power``` to I update. Will try to switch ```test_power_management_energy_over_switch``` to use 2 heater.

adi90x:
Hello, 
Was quick in fact previous test was using ```mean_power_cycle``` attribute instead of ```device_power```.
Test is OK with with the commit and NOK before.

jmcollin78:
Thank you @adi90x !

PR #200, opened 2023-11-18
Improved Support for Affordable Aliexpress Thermostats (e.g., TV02-Zigbee) [part one / basic support]

j0nxs:
Hi jmcollin78,

I hope this message finds you well. I've been actively using your project and have noticed some areas where I could contribute to enhancing the support for affordable Aliexpress thermostats, specifically the TV02-Zigbee model.

Before discovering your project, I was using better_thermostat, which generally worked well for me. However, I encountered occasional outliers where the temperature offset would get stuck at a large value (e.g., -5), causing the thermostats to attempt to heat up to 26°C. With my modifications in this project here, I've managed to achieve reliable functionality with the TV02-Zigbee devices. Although the implementation is not perfect yet, I've made substantial progress.

My plan moving forward is to continue refining the support for these thermostats. I'm also looking into leveraging native features provided by the thermostats in Home Assistant, such as native frost protection, etc. To achieve this, I'm considering creating a 'FeatureController' class for the UnderlayingClimates. This class would automatically recognize various features and pass them as child setups for the UnderlayingClimate.

Additionally, I aim to introduce a local calibration algorithm similar to that in better_thermostat. This algorithm continually adjusts the calibration to ensure that the integrated temperature sensor aligns with the external one.

I believe these enhancements will contribute to a more robust and versatile integration of affordable Aliexpress thermostats into your project. I'm eager to collaborate and receive feedback on these proposed changes.

Thank you for considering my pull request, and I look forward to your thoughts on these improvements.

Improvements for the future:
- [ ] Cleaner implementation for the heatstop switch
- [ ] Offset-Based Temperatur Controlling for the Thermostats which supports local_callibration

Here some results of the regulation with thermostats own regulation with this changes:
image
image

A small problem is that the heat state is not always reported correctly - the state is heat when the set_point > current_temperature, but these thermostats seems to heat a little bit when they are already on the temperature to hold the temperature, i am unsure how to improve that. The TV02-Zigbee doesn't report any status for heating.


maia:
@j0nxs, as for the local calibration you're suggesting, this is partly available in the "self regulation" mode, specifically in the "slow" setting, which will adjust the target temperature based on the average difference between target temperature and measured temperature over the past 24 hours. I do think this feature is too hidden and should also be named "calibration offset" instead of self-regulation, but as there's a lot of optimisations that can be done here it might be better to first think about what we need and want and how it should be configurable by the users. For example I believe there are a couple of different time frames that all are relevant and should be available – currently you have to choose between slow and fast:

1. long term offset between target temp and measured temp due to calibration offsets of the TRV
2. multiple hour effects of overheating due to solar radiation, cooking,…
3. external weather: does someone want the internal target temperature *lower* or *higher* when it's very, very cold outside?
4. quick regulation: I cannot say anything about that, but I believe it's for electrical heating and buildings that do not store heat well.

Also I think these features might need different algorithms. The long term offset might be a candidate for EMA (exponential moving average), the overheating due to external effects might be handleable with the temperature slope updates that @jmcollin78 is considering for next weekend (especially the "slow" curve that looks at the changes over the past 60 minutes), the quick regulation might need a PI-algorithm as currently used.

If you have any proposal for parts of these concepts, please post them. I find it best to first draft a concept together, otherwise the code and configuration will get very messy over time due to all these features.

jmcollin78:
Hello @j0nxs,

Thank you so much for this PR.

For regulation, I'm totally aligned with @maia. A new feature called self-regulation have been introduced in 3.8.0 release and is now almost mature. A new release with important improvements have been release yesterday.

I am confident that with this and @maia improvements we will have something which is working fine without having to do more than this.

You can have  at a look at the release 4.1.0 releases yesterday and try the 'Strong regulation'. I'm confident that will solve your issue. Else of course we can help to tune the parameters.

Concerning the 'FeatureController' class for the UnderlayingClimates, I will have a look as this could be an interesting way to add features with clear separation of concern in the code.

jmcollin78:
Hello @j0nxs ,

Do you have tried the regulation mode ? It is now fully operational. It would be great if you could confirm it is working on your environement.

If yes I will close this PR because it is no more needed. Thanks anyway for this PR.

CodingGhost:
I just want to add, that the self regulation does not replace the local calibration. the PI algorithm cant work correctly if the TRV thinks it has a different temperature than what Versatile thermostat thinks the current temperature is. I have had this exact problem with VT, the VT regulation creates a +2degree offset because its way to cold, but the TRV thinks its already 2degree warmer so it still doesnt heat. I solved it by creating my own local calibration automation, which greatly improved the heating accuracy.

maia:
> I just want to add, that the self regulation does not replace the local calibration. the PI algorithm cant work correctly if the TRV thinks it has a different temperature than what Versatile thermostat thinks the current temperature is. I have had this exact problem with VT, the VT regulation creates a +2degree offset because its way to cold, but the TRV thinks its already 2degree warmer so it still doesnt heat. I solved it by creating my own local calibration automation, which greatly improved the heating accuracy.

I've also had this issue and the current logic does not allow to implement a (long term) calibration offset. I started a discussion in #208 and would love to see your comments in that discussion, ideally also the automations you created, are you willing to share it? Thanks!

jmcollin78:
> the PI algorithm cant work correctly if the TRV thinks it has a different temperature than what Versatile thermostat thinks the current temperature is

On the contrary I believe this will work with a Strong regulation. At each cycle, the error factor is incremented and so contribute to a self calculated offset. So if 2° is not enough, try with the strong regulation and you will up to 5° of regulation. This error is now persistent (thanks to @maia idea) so you have a dynamically adaptable offset with this solution.

Can you please give it a try ? You have to wait let's say 12h to have a correct initialisation.

PR #227, opened 2023-11-23
Add Greek language translation

vassilis-papanikolaou:
Adding "el.json" file for Greek translations of the user strings.

Note:
"el" is the ISO 639-1 standard two-letter code for Greek language.

Source:
https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes#el


jmcollin78 (approved):

jmcollin78:
Thank you @vassilis-papanikolaou !

PR #345, opened 2024-01-17
Implement a keep-alive feature for directly controlled heater switches

pdcastro:
The heater switch keep-alive feature consists of regularly refreshing the state of directly controlled switches at a configurable interval (regularly turning the switch ‘on’ or ‘off’ _again,_ even if it is already turned ‘on’ or ‘off’), just like the `keep_alive` setting of Home Assistant’s Generic Thermostat integration:

> HASS Generic Thermostat https://www.home-assistant.io/integrations/generic_thermostat/
> `keep_alive` _integer (optional)_
> _Set a keep-alive interval. If set, the switch specified in the heater option will be triggered every time the interval elapses. Use with heaters and A/C units that shut off if they don’t receive a signal from their remote for a while. Use also with switches that might lose state._


pdcastro:
@jmcollin78, this PR proposes the implementation of a keep-alive feature as described. If you agree to having such a feature added as proposed in this PR generally, I can then work further on it, especially adding documentation and translations.

When rebased on tag `5.3.3.beta3`, all tests pass in the VSCode dev container. I have also tested “the real thing” with an appliance of mine and Home Assistant version 2024.1.3. The code was linted with black.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -202,6 +204,7 @@ def __init__(
         self._on_time_sec = 0
         self._off_time_sec = 0
         self._hvac_mode = None
+        self._keep_alive = IntervalCaller(hass, keep_alive_sec)
```
pdcastro, line 207: Note: To simplify the code, I chose to always instantiate this object even if `keep_alive_sec` is unset or zero, but in this case it will do nothing and the feature will be effectively disabled. I tested it when `keep_alive_sec` is unset, zero, and non-zero.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -214,6 +217,13 @@ def is_inversed(self):
         """Tells if the switch command should be inversed"""
         return self._thermostat.is_inversed
 
+    @overrides
+    def startup(self):
+        super().startup()
+        self._keep_alive.set_async_action(
+            self.turn_on if self.is_device_active else self.turn_off
+        )
```
pdcastro, line 225: As per other comment, even though this piece of code always runs, it will do nothing if the `keep_alive` setting is unset or zero.

jmcollin78:
Oooh I will see that. Thank you for this PR !

For over_switch, because of the cycle which constantly turns on and off, I guess there is no need of keep_alive (that is why I don't implement it initialy). And don't know for over_climate if this make sens, because the underlying climate may do it alone and may be for `over_valve` ?

I'm not sure this is useful finaly. In what case do you need this feature ?

pdcastro:
> Oooh I will see that. Thank you for this PR !
> 
> For over_switch, because of the cycle which constantly turns on and off, I guess there is no need of keep_alive (that is why I don't implement it initialy). And don't know for over_climate if this make sens, because the underlying climate may do it alone and may be for `over_valve` ?
> 
> I'm not sure this is useful finaly. In what case do you need this feature ?

Thanks for the feedback. The keep-alive feature is intended for “over_switch” operation — directly controlled switch entities. When a TPI cycle is delivering high power, say 100% (when the target temperature is well above the current temperature and the “on_percent” variable is near 1 = 100%), I understand that the switch is turned ‘on’ at the beginning of the cycle, and no further action (switch on/off) is taken on the switch until the end of the cycle. Does this match your understanding as well? In this case, if the appliance requires a keep-alive interval that is shorter than the TPI cycle, then the appliance will switch off (failsafe mode) before the end of the TPI cycle, delivering less power than intended.

I have definitely observed this behaviour with a small electric heater that requires a short keep-alive interval of 20 seconds, with a TPI cycle of 1 minute and intended 100% power. Without the keep-alive feature implemented in this PR, the appliance would switch off about 20 seconds into the TPI cycle, thus delivering 33% power (20 seconds ‘on’, 40 seconds ‘off’) even when the intended TPI power was 100% (60 seconds 'on', 0 seconds 'off').

Reducing the TPI cycle duration to match the required keep-alive interval may work in some cases, but not all cases. A larger heating system based on oil or natural gas (currently over three quarters of homes here in the UK, for example) may require longer TPI cycles such as 10 or 20 minutes,* while switch keep-alive intervals might be 1 or 5 minutes (or any other value, depending on the system design).

\* Example: My central heating system takes up to a minute between heat demand being presented and the boiler igniting the flame, so a TPI cycle of 1 minute would just not work. When heat demand is presented to the system, a mechanical water valve slowly opens and some 20 seconds later, a water pump turns on heat demand is presented to a natural gas boiler. The boiler then turns on a fan and takes another 20 to 40 seconds to ignite the main flame. I find that 20 minutes is a reasonable TPI cycle for this system.

jmcollin78:
> When a TPI cycle is delivering high power, say 100% (when the target temperature is well above the current temperature and the “on_percent” variable is near 1 = 100%), I understand that the switch is turned ‘on’ at the beginning of the cycle, and no further action (switch on/off) is taken on the switch until the end of the cycle

Yes, you are totally right. When 100% or 0% the switch is not switched but only at start of the first cycle. THe use case is now clear for me. I will have a look before going further to the development of the release (based on beta already pubished).

custom_components/versatile_thermostat/thermostat_switch.py
```diff
@@ -101,6 +102,7 @@ def post_init(self, config_entry):
                     thermostat=self,
                     switch_entity_id=switch,
                     initial_delay_sec=idx * delta_cycle,
+                    keep_alive_sec=config_entry.get(CONF_HEATER_KEEP_ALIVE),
```
jmcollin78: kee_alive_sec is an int. With this it is possible to send None (even if the default value is 0, for existing VTherm which doesn't have the option nor the default value. This is a common trap).
So you should send 0 if None.

custom_components/versatile_thermostat/thermostat_switch.py
```diff
@@ -121,6 +123,7 @@ async def async_added_to_hass(self):
                     self.hass, [switch.entity_id], self._async_switch_changed
                 )
             )
+            switch.startup()
```
jmcollin78, line 130: Why this ? Startup will be done after all configuration is done in the async_control_heating.

I think this is not necessary and is in doublon.

custom_components/versatile_thermostat/keep_alive.py
```diff
@@ -0,0 +1,43 @@
...
+        self._interval_sec = interval_sec
+        self._remove_handle: CALLBACK_TYPE | None = None
+
+    def cancel(self):
+        if self._remove_handle:
+            self._remove_handle()
+            self._remove_handle = None
+
+    def set_async_action(self, action: Callable[[], Awaitable[None]]):
+        if not self._interval_sec:
```
jmcollin78, line 40: I prefer testing explicitly 0 or None because both values are possible.

custom_components/versatile_thermostat/keep_alive.py
```diff
@@ -0,0 +1,43 @@
...
+            return
+        self.cancel()
```
jmcollin78, line 42: You should cancel if 0 also. May be first instruction is cancel, and then do the job.

custom_components/versatile_thermostat/underlyings.py
jmcollin78: At each cycle, the heater turns on and then off. If the keep_alive is programmed with 'turn_on' and the heater is in the off phase, this will turns it on. 
I guess there is a big conception mistake here.

You should rather initialize the keep_alive at each turn_on, turn_off call, I guess.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -240,6 +250,7 @@ def is_device_active(self):
     # @overrides this breaks some unit tests TypeError: object MagicMock can't be used in 'await' expression
     async def turn_off(self):
         """Turn heater toggleable device off."""
+        self._keep_alive.cancel()
```
jmcollin78: Not necessary because set_async_action already cancels

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -254,8 +265,11 @@ async def turn_off(self):
         except ServiceNotFound as err:
             _LOGGER.error(err)
 
+        self._keep_alive.set_async_action(self.turn_off)
```
jmcollin78: Should be after async_call to avoid setting a keep_alive if service have not been called.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -254,8 +265,11 @@ async def turn_off(self):
...
+
     async def turn_on(self):
         """Turn heater toggleable device on."""
+        self._keep_alive.cancel()
```
jmcollin78: Not necessary as above

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -269,6 +283,8 @@ async def turn_on(self):
         except ServiceNotFound as err:
             _LOGGER.error(err)
 
+        self._keep_alive.set_async_action(self.turn_on)
```
jmcollin78: To put in between the try / except

tests/test_start.py
```diff
@@ -264,6 +264,7 @@ async def test_over_switch_deactivate_preset(
             CONF_HEATER_2: None,
             CONF_HEATER_3: None,
             CONF_HEATER_4: None,
+            CONF_HEATER_KEEP_ALIVE: 0,
```
jmcollin78, line 267: Please add a unit test with KEEP_ALIVE not 0 to see if method are called.

You can inspire from the existing tests `test_bug_66` which mocks method, check they are called and waits for the window to be open/close.

Let me know if it is too complex I'm pretty sure I can help.

custom_components/versatile_thermostat/underlyings.py
jmcollin78: I  have seen after that it is done below. You can forget this comment.

jmcollin78 (commented):
Great improvement @pdcastro !

I will take this PR with pleasure. The comments are easy to fix, I guess.

custom_components/versatile_thermostat/underlyings.py
pdcastro: OK. As you know, this piece of code only runs when the integration is loaded or reloaded — at start up time. (See my other comment for the reason why this is needed at start up time.) What I meant for this piece of code to do is to start sending keep-alive messages† as soon as in the integration is loaded — “turn on” messages if the switch is initially turned on, or “turn off” messages if the switch is initially turned off. See another comment for a scenario where this is required. From my testing, it works well.
† I use the word “messages” because I am working with MQTT switches.

custom_components/versatile_thermostat/thermostat_switch.py
pdcastro: Two related matters here:

1. _About the `UnderlyingSwitch.startup()` method_
The `startup()` method is empty in the `UnderlyingEntity` base class. It was overridden in the `UnderlyingClimate` subclass, but not in the `UnderlyingSwitch` or the `UnderlyingValve` subclasses. For `UnderlyingClimate`, the `startup()` method is called through the chain `BaseThermostat.async_added_to_hass()` ➔ `BaseThermostat.async_startup()` ➔ `ThermostatOverClimate.init_underlyings()`. However, `ThermostatOverSwitch` does not override `init_underlyings()` and does not call `UnderlyingSwitch.startup()` (before this PR), so this call is not a duplicate in this sense. By the way, I considered overriding `init_underlyings()` in `ThermostatOverSwitch`, but for the purposed of keep-alive messages / service calls, I wanted it to happen as late as possible in the initialisation of the integration, and `init_underlyings()` is actually called somewhat early (relatively). So I chose this solution, calling `UnderlyingSwitch.startup()` towards the end of `ThermostatOverSwitch.async_added_to_hass()`.
To confirm, if this line of code is commented out, `UnderlyingSwitch.startup()` is never called.

2. _What happens when the integration is loaded or reloaded_
I think this relates to your point when you mention `async_control_heating()`. If Home Assistant is (re)started or the Versatile Thermostat integration is (re)loaded and the thermostat is initially switched off (`HVACMode.OFF`) and the switch is already turned off (`is_device_active` is False), then the `UnderlyingSwitch` `turn_on` and `turn_off` methods are not called at all. In this scenario, without explicit initialisation through some such `startup()` method, regular keep-alive messages / service calls are not started until the user turns on the thermostat. But why are keep-alive messages needed in this scenario, anyway? Some reasons:
   - Heating appliances that require keep-alive messages usually make a distinction between “turn off” due to an explicit turn off message, which is normal operation, and “turn off” due to missing keep-alive messages, which is an error condition. It would be a better user experience if the appliance ceased reporting an error condition as soon as the integration was loaded (if correctly configured), even if the thermostat was turned off (`HVACMode.OFF`).
   - While the thermostat is an active use, say mid winter, it may not seem a big deal to expect the user to turn on the thermostat for keep alive to be triggered, however at a time when the thermostat is turned off, say mid summer, if the user restarts Home Assistant for whatever reason (other integrations), turn off keep alive messages should be resumed without requiring the user to turn on the thermostat.
   - Without this `startup()` call (or alternative solution to a similar effect), we would have to document something along the lines of _“Please note that keep-alive messages will only start after the thermostat is first switched on,”_ but this would sound like a deficiency that needs working around by the user.
   - Less important but still a reason, there would be an “inconsistent state” corner-case scenario where the appliance is currently turned on, the keep-alive interval is relatively long (say, 10 minutes), the integration is “reset” or reinstalled and starts with `HVACMode.OFF`, but the appliance remains turned on for several minutes until its failsafe mode kicks in.

custom_components/versatile_thermostat/underlyings.py
pdcastro: I had wanted to prevent a potential imminent keep-alive message (MQTT) being triggered during the async context switch in the middle of this method (`await self._hass.services.async_call(...)`), before the call to `set_async_action()` that “updates” the keep-alive messages (from ‘on’ to ‘off’ or vice-versa), especially if Home Assistant was running a on a resource-stretched low-power device such as a Raspberry Pi 2 where the async context switch could take longer. However, on second thoughts, I agree that this optimisation is not sufficiently justified and I will delete this line.

custom_components/versatile_thermostat/underlyings.py
pdcastro: Good point. 👍 In fact, if the service call raises any exception at any time for any reason, I think keep alive should be canceled too, otherwise it would be repeatedly raising the same errors. I am changing the code to achieve this too.

custom_components/versatile_thermostat/keep_alive.py
pdcastro: This connects to another comment where you had pointed out that `None` could be introduced by `config_entry.get(CONF_HEATER_KEEP_ALIVE)`. However, the type hint for `interval_sec` in the `__init__` signature is `int`, so rather than mistrusting the type hint or amending it to be `int | None`, I will add some validation code up the call chain to make sure that the type hint is respected and the `None` value is really not possible here.

custom_components/versatile_thermostat/keep_alive.py
pdcastro: If the user configured `keep_alive_sec` as 0 (or unset or `None`), then there is nothing to cancel because no callback is setup in the first place. This can be “proven” by looking at the `IntervalCaller` class in isolation, because `set_async_action()` is the only method that assigns a callback to `self._remove_handle`, and it can only happen if `keep_alive_sec` is non zero. Also, if the user reconfigures the integration (changing `keep_alive_sec` from zero to non-zero or vice-versa) while Home Assistant is running, then 1- `IntervalCaller.cancel()` is called by `remove_entity()`, and 2- a new instance of `IntervalCaller` is created. So in any of these scenarios, I reckon that there’s no need to call `cancel` when `keep_alive_sec` is 0 (or unset or `None`).

custom_components/versatile_thermostat/thermostat_switch.py
pdcastro: There is a “positive int” constraint in the config flow code and also a default value of zero, in case the option is not set by the user:
```python
vol.Optional(CONF_HEATER_KEEP_ALIVE, default=0): cv.positive_int,
```
But it sounds like this is not enough, so I will add some enforcement code to be sure, e.g.:
```python
keep_alive_sec = max(0, int(config_entry.get(CONF_HEATER_KEEP_ALIVE) or 0))
```

tests/test_start.py
pdcastro: Done 👍

pdcastro (commented):
Thanks for you review. 👍  I am adding new commits that address the points that were raised, to make it easier for you to see what has changed. Before merging the PR though, I could squash the commits to make the history tidier.
I also still need to add documentation and translations (Google Translate...). I can add those to this same PR, or I can create a separate PR. Let me know if you have a preference.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -245,12 +255,13 @@ async def turn_off(self):
...
-                command,
-                data,
-            )
+            try:
+                data = {ATTR_ENTITY_ID: self._entity_id}
+                await self._hass.services.async_call(domain, command, data)
+                self._keep_alive.set_async_action(self.turn_off)
+            except Exception:
+                self._keep_alive.cancel()
+                raise
```
pdcastro, line 264: Building on previous comments, I’ve added an inner `try` block that catches and re-raises any exception for the purpose of cancelling keep alive if the service call fails for any reason. I had the following cases in mind:
- Keep alive initiated through the `startup()` method, before it is known whether the service calls will succeed.
- Keep alive initiated here, such that at least one service call succeeded, but where subsequent service calls might fail (for any unforeseen reason, perhaps in case the user renames or removes the service while Versatile Thermostat is running).

Adding the inner try/except block allows a single `cancel()` call to be made and avoids `isinstance` tests, as in the following alternatives:

```python
        try:
            data = {ATTR_ENTITY_ID: self._entity_id}
            await self._hass.services.async_call(domain, command, data)
            self._keep_alive.set_async_action(self.turn_on)
        except ServiceNotFound as err:
            _LOGGER.error(err)
            self._keep_alive.cancel()
        except Exception:
            self._keep_alive.cancel()
            raise
```

Or:
```python
        try:
            data = {ATTR_ENTITY_ID: self._entity_id}
            await self._hass.services.async_call(domain, command, data)
            self._keep_alive.set_async_action(self.turn_on)
        except Exception as e:  # pylint: disable=broad-exception-caught
            self._keep_alive.cancel()
            if isinstance(e, ServiceNotFound):
                _LOGGER.error(e)
            else:
                raise
```

By the way:
- `pylint` doesn’t seem to complain about “broad exception caught” when the exception is unconditionally re-raised, which is why I did not need to add `# pylint: disable=broad-exception-caught` in some of the alternatives.
- `black` was happy with the re-formatting.

custom_components/versatile_thermostat/keep_alive.py
```diff
@@ -0,0 +1,53 @@
...
+
+        async def callback(_time: datetime):
+            try:
+                await action()
+            except Exception as e:  # pylint: disable=broad-exception-caught
+                _LOGGER.error(e)
+                self.cancel()
```
pdcastro, line 49: The added try/except block avoids ugly job/task stack traces printed in the Home Assistant logs, which would look like a bug in the Versatile Thermostat integration. For example, a raising an unexpected “foo” exception _without_ the try/except block:
```
2024-01-23 23:50:21.545 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved:   File "/home/vscode/.local/bin/hass", line 8, in <module>
    sys.exit(main())
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/__main__.py", line 209, in main
    exit_code = runner.run(runtime_conf)
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/runner.py", line 188, in run
    return loop.run_until_complete(setup_and_run_hass(runtime_config))
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 640, in run_until_complete
    self.run_forever()
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 607, in run_forever
    self._run_once()
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 1914, in _run_once
    handle._run()
  File "/usr/local/lib/python3.11/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/helpers/event.py", line 1454, in _run_async_call_action
    hass.async_run_hass_job(job, time_tracker_utcnow())
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/core.py", line 671, in async_run_hass_job
    hassjob.target(*args)
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/helpers/event.py", line 1532, in interval_listener
    hass.async_run_hass_job(job, now)
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/core.py", line 674, in async_run_hass_job
    return self.async_add_hass_job(hassjob, *args)
  File "/home/vscode/.local/lib/python3.11/site-packages/homeassistant/core.py", line 568, in async_add_hass_job
    task = self.loop.create_task(hassjob.target(*args), name=hassjob.name)
  File "/workspaces/versatile_thermostat/custom_components/versatile_thermostat/keep_alive.py", line 47, in callback
    raise Exception("foo")
Exception: foo
```
And _with_ the try/except block:
```
2024-01-23 23:49:32.513 ERROR (MainThread) [custom_components.versatile_thermostat.keep_alive] foo
```

pdcastro (commented):
New commits added to address the review comments. To confirm:
- All tests pass (`pytest` in the VS Code dev container).
- Code formatted with `black`.
- Code checked with `pylint` and `pylance`.

custom_components/versatile_thermostat/thermostat_switch.py
pdcastro: Actually, if it is just `None` that we are guarding against, the following should be enough, no?
```python
UnderlyingSwitch(
    ... ,
    keep_alive_sec=config_entry.get(CONF_HEATER_KEEP_ALIVE, 0),
)
```
I did some testing and I can confirm that:
* `config_entry.get(SOME_UNSET_ENTRY, 0)` returns `0` rather than `None`.
* When using the Home Assistant integration configuration user interface (config flow), the UI prevents me from entering negative numbers and decimal points, and if I delete the contents of the form field (unsetting it), `config_entry.get(SOME_UNSET_ENTRY)` actually returns `0` rather than `None`, thanks to the schema constraint in `config_schema.py`:

```python
vol.Optional(CONF_HEATER_KEEP_ALIVE, default=0): cv.positive_int,
```

jmcollin78: I would do that:
```
keep_alive_sec=config_entry.get(CONF_HEATER_KEEP_ALIVE) or 0 
```
but 
```
config_entry.get(CONF_HEATER_KEEP_ALIVE, 0)
```
(which I don't was aware until now), seems good also.

custom_components/versatile_thermostat/thermostat_switch.py
jmcollin78: ok. Thank for the explanation.

> “Please note that keep-alive messages will only start after the thermostat is first switched on,”

So, if I understand well, in summer, when Vtherm is off, the keep-alive will continue to be send to the device ? Am I wrong ? Why not but I wonder how to avoid that. I consider permanently running features even if turned off is a bad pattern. But maybe I understand a wrong way.

custom_components/versatile_thermostat/keep_alive.py
jmcollin78: Yes good point. Testing once for all before arriving here is better. I confirm.

---
In a general way, an int can always be None. Python is not a strongly typed language (like Javascript if you want). Typing is more for an indication for helping the user.

custom_components/versatile_thermostat/keep_alive.py
jmcollin78: Yes ... but ... for the future robustness of the code (many people as me will change this in the future) is it, in my opinion better to cancel() even if you think there is nothing to cancel.

This kind of programing "save my life" many time as I handles also non nominal cases. There is many piece of code like this (stop heater even if "normally" it is already turned off -> this is safety). Testing if xx <= 0 instead of if xxx == 0, ... all this patterns are for safety and robustness.

In that case, in the cancel method, if really there is nothing to cancel, then you do nothing. So, the cancel() method have to be think as "cancel_if_any", if you prefer.

jmcollin78:
> Thanks for you review. 👍 I am adding new commits that address the points that were raised, to make it easier for you to see what has changed. Before merging the PR though, I could squash the commits to make the history tidier. I also still need to add documentation and translations (Google Translate...). I can add those to this same PR, or I can create a separate PR. Let me know if you have a preference.

I prefer to have all in the same PR.

custom_components/versatile_thermostat/keep_alive.py
pdcastro: My design choice of always calling `self._keep_alive.set_async_action(...)` in the `turn_on` and `turn_off` methods, in order to simplify the code, already means that this method is always called even when the keep alive feature is not used — which is indicated by `interval_sec` being zero. This is why I put that test at the very beginning of `set_async_action()` — to waste no more CPU than has already been wasted. You are asking me to additionally call `cancel()` even when the keep alive feature is disabled (`interval_sec` is zero), in the hope that somehow this may prevent someone else from introducing a bug in the future. I will do it if you insist further! 🙂

custom_components/versatile_thermostat/thermostat_switch.py
pdcastro: “VTherm” is short for “Versatile Thermostat”, right? I ask in order to be sure that we are on the same page. 🙂 (A web search pointed me to [something else](https://www.vitalenergi.co.uk/products-services/energy-supply-consumption-management/glass-smart-app/vtherm-e-glass/)!)
And I assume that “when VTherm is off” means “when `self._hvac_mode == HVACMode.OFF`”.

This is a design choice that we can make, and here's the main argument. A heating appliance that relies on keep-alive messages (think MQTT) may display an error code to the user (perhaps on a front-panel LCD display) if the keep-alive messages are missing, whether they are “turn on” or “turn off” messages. If we stopped keep alive messages from being sent when `self._hvac_mode == HVACMode.OFF`, the appliance would display such an error code. I thought we would want to avoid such an error code as long as VTherm is configured to produce keep-alive messages, even if hvac mode is ‘off’.

Picture also this “social” situation: Mary, a technically minded person, sets up Home Assistant + VTherm to control a heating appliance for Bob, who is not as technically minded. Bob then uses the VTherm UI card on a phone to control heating at home. Summer arrives and Bob sets hvac mode to 'off', blissfully unaware of technical implications such as keep alive messages. _A week later,_ by chance, Bob happens to glance at the appliance's front panel and sees an error code that he does not understand. He immediately assumes that something is _broken,_ and he may blame Mary for setting up something complicated that goes wrong! Then Mary has to investigate what is going on, as she may herself have forgotten most of the details of what she had configured a few months earlier.

If a technically minded user specifically wanted to stop keep alive messages over an extended period of not using the appliance (e.g., summer), they could set the keep alive interval to zero in the configuration. In any case, we would document that setting the keep alive interval to zero disables the feature.

Something I think we should ***not*** do is to add yet another configuration option for the user to choose whether or not keep alive messages should be sent while hvac mode is ‘off’. There are already too many configuration options! We should make this choice for the user, one way or another.

custom_components/versatile_thermostat/thermostat_switch.py
pdcastro: _**Edit:** The correct type of `ConfigEntry.data` is `MappingProxyType`, not `Mapping`._

`config_entry` has the type of [ConfigEntry.data](https://github.com/home-assistant/core/blob/2024.1.5/homeassistant/config_entries.py#L267), which is [types.MappingProxyType](https://docs.python.org/3.11/library/types.html#types.MappingProxyType), which is a kind of read-only dictionary, more precisely a read-only view over a dictionary (if the original dict changes, the view reflects those changes, but the dict cannot be changed through the view) –

```
$ python
Python 3.11.7 (main, Dec 19 2023, 17:11:23) [GCC 10.2.1 20210110] on linux

>>> from types import MappingProxyType
>>> d = {'a': 1}
>>> e = MappingProxyType(d)
>>> d['a']
1
>>> e['a']
1
>>> d['a'] = 2
>>> e['a']
2
>>> e['a'] = 3
  File "<stdin>", line 1, in <module>
TypeError: 'mappingproxy' object does not support item assignment
>>> d is e
False
>>> d == e
True
>>> f = e.copy()
>>> e is f
False
>>> e == f
True
>>> d == f
True
>>> d.__class__.__name__
'dict'
>>> e.__class__.__name__
'mappingproxy'
>>> f.__class__.__name__
'dict'
```

While investigating this, I got a bit carried away and created a separate PR that adds more type hints in a few files: #364

custom_components/versatile_thermostat/thermostat_switch.py
jmcollin78: > “VTherm” is short for “Versatile Thermostat”, right?

Yes.

Ok with your message.

README-fr.md
```diff
@@ -244,7 +244,10 @@ En fonction de votre choix sur le type de thermostat, vous devrez choisir une ou
 > 2. **quelle type de régulation je veux ?** Si l'équipement piloté possède son propre mécanisme de régulation (clim, certaine vanne TRV) et que cette régulation fonctionne bien, optez pour un ```over_climate```
 
 ### Pour un thermostat de type ```thermostat_over_switch```
-![image](https://github.com/jmcollin78/versatile_thermostat/blob/main/images/config-linked-entity.png?raw=true)
+![image](images/config-linked-entity.png)
+
+Certains switches de radiateur nécessitent des « messages keep-alive » pour les empêcher de déclencher un arrêt de sécurité. Ceci peut être activé en entrant un nombre de secondes non nul dans le champ d'intervalle keep-alive du switch. En cas de doute, laissez-le vide ou zéro (valeur par défaut).
```
pdcastro: Pardon my French! And blame Google Translate for it. 🙂

custom_components/versatile_thermostat/translations/fr.json
```diff
@@ -46,6 +46,7 @@
           "heater_entity2_id": "2ème radiateur",
           "heater_entity3_id": "3ème radiateur",
           "heater_entity4_id": "4ème radiateur",
+          "heater_keep_alive": "Intervalle keep-alive du switch en secondes",
```
pdcastro, line 49: I saw that the word “switch” was already being used elsewhere in the French translation, so I used it here too instead of “interrupteur”, but you tell me! As for “keep-alive”, I wonder if it is a technical term that is more readily understood in English than when translated. Again, you tell me. Google Translate actually suggested some strange things equivalent to _“Change the interval of life maintenance”._ 😖

custom_components/versatile_thermostat/translations/el.json
```diff
@@ -50,6 +51,7 @@
           "heater_entity2_id": "Προαιρετική 2η ταυτότητα οντότητας θερμαντήρα. Αφήστε κενό αν δεν χρησιμοποιείται",
           "heater_entity3_id": "Προαιρετική 3η ταυτότητα οντότητας θερμαντήρα. Αφήστε κενό αν δεν χρησιμοποιείται",
           "heater_entity4_id": "Προαιρετική 4η ταυτότητα οντότητας θερμαντήρα. Αφήστε κενό αν δεν χρησιμοποιείται",
+          "heater_keep_alive": "Optional heater switch state refresh interval. Disabled if empty or zero",
```
pdcastro: I had a go at French and Italian because I had some hope of being able to make absurd Google translations slightly less absurd, but Greek and Slovak I will not even try. A copy-and-paste of Google Translate would end up as some nonsense like _“Option to replace the country’s freshen up intermission”_ (or worse).

So I entered English as placeholder for the Greek and Slovak translations. I suggest that we merge this PR (when you feel like it is tolerable), and straight after that we invite **`misa1515`, `vassilis-papanikolaou`** and **`Bergoglio`** to provide translations (either they create a new PR, or we do). We could also put this PR on hold until they provide translations, but this PR already has 40+ comments, touches 21 files, and if they took a while to reply, the PR could end up with merge conflicts.

custom_components/versatile_thermostat/keep_alive.py
pdcastro: > I will do it if you insist further!

You haven’t yet insisted further, 🫣 so I am leaving it as it is. I understand that you are not quite happy with it, but I think I can live with that! 😄😇 The code is good.

pdcastro:
PR status: I have added the documentation and translations for review, with some comments. Other than that, I think the PR is ready for merging. I have squashed two of the commits too (tidier history). I have not marked any conversations as “resolved” — I leave that to you as the reviewer. If you think I have missed any conversation, please point it out to me. Thanks!

custom_components/versatile_thermostat/translations/el.json
jmcollin78: No. HA will do itself the fallback to the En version. Just don't add the values for language you don't know.

Some contributors make a PR regularly to adapt the text. I will ping them after the release so that they will be able to add their texte.

README-fr.md
jmcollin78: Really good ! But to be more compliant with the rest of the article, please replace with:

Certains équipements nécessitent d'être périodiquement sollicités pour les empêcher un arrêt de sécurité. Connu sous le nom de "keep-alive" cette fonction est activable en entrant un nombre de secondes non nul dans le champ d'intervalle keep-alive du thermostat. Pour désactiver la fonction ou en cas de doute, laissez-le vide ou entrez zéro (valeur par défaut).

custom_components/versatile_thermostat/keep_alive.py
jmcollin78: yes. That is more a matter of programming style, and I have to accept other programming style (or reject all PR and I will stay alone with the code all my life) 😉

custom_components/versatile_thermostat/keep_alive.py
jmcollin78: Yes. Never let stacktrace on HA logs is a good practice.

custom_components/versatile_thermostat/translations/fr.json
```diff
@@ -67,6 +68,7 @@
           "heater_entity2_id": "Optionnel entity id du 2ème radiateur",
           "heater_entity3_id": "Optionnel entity id du 3ème radiateur",
           "heater_entity4_id": "Optionnel entity id du 4ème radiateur",
+          "heater_keep_alive": "Optionnel intervalle de rafraîchissement du switch. Désactivé si vide ou zéro",
```
jmcollin78: "Intervalle de rafraichissement du switch en secondes. Mettre 0 ou vide pour désactiver. À n'utiliser que pour les switchs qui le nécessite."

custom_components/versatile_thermostat/translations/en.json
```diff
@@ -302,6 +305,7 @@
           "heater_entity2_id": "Optional 2nd Heater entity id. Leave empty if not used",
           "heater_entity3_id": "Optional 3rd Heater entity id. Leave empty if not used",
           "heater_entity4_id": "Optional 4th Heater entity id. Leave empty if not used",
+          "heater_keep_alive": "Optional heater switch state refresh interval. Disabled if empty or zero",
```
jmcollin78: add "Use it only if your switch need it." at the end of the description. Idem for step/option description.

custom_components/versatile_thermostat/translations/sk.json
```diff
@@ -45,6 +45,7 @@
           "heater_entity2_id": "2. spínač ohrievača",
           "heater_entity3_id": "3. spínač ohrievača",
           "heater_entity4_id": "4. spínač ohrievača",
+          "heater_keep_alive": "Switch keep-alive interval in seconds",
```
jmcollin78: remove. Fallback is done by HA itself. Idem below. If you are sure for Italian, you can keep it.

tests/test_switch_keep_alive.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Every time the keep-alive callback is called, the home assistant switch
+        # turn on/off service should be called too.
+        self._assert_service_call(
+            common_mocks,
+            [
+                call("switch", SERVICE_TURN_OFF, {"entity_id": "switch.mock_switch"}),
+                call("switch", SERVICE_TURN_OFF, {"entity_id": "switch.mock_switch"}),
+            ],
+        )
```
jmcollin78, line 256: Great test class ! I have learn something. Really nice.

jmcollin78:
We are almost done. And you have now a conflict in readme since your PR in images. Sorry for this but it should be easy to fix.

This is really a great job you have done 😍

custom_components/versatile_thermostat/translations/en.json
pdcastro: I agree that it makes sense to advise the user that this feature should only be used when required, 👍 as it has “performance implications” of sorts. But I also noticed that the whole description was getting long:

```python
"heater_keep_alive": "Optional heater switch state refresh interval. Disabled if empty or zero. Use it only if your switch needs it."
```

 The preceding descriptions, for example, have a more succinct _“Leave empty if not used”_ instruction:
```python
"heater_entity2_id": "Optional 2nd Heater entity id. Leave empty if not used",
"heater_entity3_id": "Optional 3rd Heater entity id. Leave empty if not used",
"heater_entity4_id": "Optional 4th Heater entity id. Leave empty if not used",
```

I had added _"Disabled if empty or zero”_ because my schema definition meant that the form was pre-filled with `0`, so the user needed to be advised of the meaning of `0`:

```python
vol.Optional(CONF_HEATER_KEEP_ALIVE, default=0): cv.positive_int,
```

But from the end user’s point of view, leaving the form empty is more straightforward than entering zero. So I've changed the schema for the default to be an empty field:

```python
vol.Optional(CONF_HEATER_KEEP_ALIVE): cv.positive_int,
```

I have tested to confirm than an empty field doe not cause any validation issues with the `positive_int` constraint, and the Python code replaces `None` with `0` to satisfy the `int` type hint:

```python
 keep_alive_sec=config_entry.get(CONF_HEATER_KEEP_ALIVE, 0)
```

So, there’s no longer a need to mention “zero” to the user. My new proposal is to say _“Leave empty if not required”:_

```python
"heater_keep_alive": "Optional heater switch state refresh interval. Leave empty if not required.",
```

I think this also removes the need to say _”Use it only if your switch needs it.”_ The preceding `heater_entity` descriptions read _”Leave empty **if not used**”_ rather than _”**if not required**”_ and I think that’s appropriate: If you are not _”using”_ an extra switch entity then you leave the field empty _”if not used”,_ whereas if your switch does not _”require”_ keep-alive messages, then you leave the field empty _”if not required”._ Furthermore, while _”Disabled if empty or zero”_ was just a piece of information, _“Leave empty if not required”_ is a more forceful instruction of what the user should do.

Would you agree? If you agree, please let me have amended French translations as well. I will wait for your feedback before updating the screenshots.

custom_components/versatile_thermostat/config_schema.py
```diff
@@ -98,6 +98,7 @@
         vol.Optional(CONF_HEATER_4): selector.EntitySelector(
             selector.EntitySelectorConfig(domain=[SWITCH_DOMAIN, INPUT_BOOLEAN_DOMAIN]),
         ),
+        vol.Optional(CONF_HEATER_KEEP_ALIVE): cv.positive_int,
```
pdcastro, line 101: Amended to remove `default=0` as explained in https://github.com/jmcollin78/versatile_thermostat/pull/345#discussion_r1468544063

custom_components/versatile_thermostat/strings.json
```diff
@@ -67,6 +68,7 @@
           "heater_entity2_id": "Optional 2nd Heater entity id. Leave empty if not used",
           "heater_entity3_id": "Optional 3rd Heater entity id. Leave empty if not used",
           "heater_entity4_id": "Optional 4th Heater entity id. Leave empty if not used",
+          "heater_keep_alive": "Optional heater switch state refresh interval. Leave empty if not required.",
```
pdcastro, line 71: See comment https://github.com/jmcollin78/versatile_thermostat/pull/345#discussion_r1468544063.
Also, regarding the final period punctuation `’.’` at the end of _”Leave empty if not required.”_ , I have noticed that it is not consistent with the preceding descriptions that to not include the final period, but I included it because:
* You included the period in your suggested English and French revisions, and
* I think including the period is the right thing to do when the sentence is “full” rather than just a “label.” For example, _”Optional heater switch”_ is just a label, whereas _“This is an optional heater switch.”_ and _”Leave empty if not required.”_ are full sentences. (And if a sentence follows a label, a period is needed at the end of the label in order to separate it from the sentence.)

README.md
```diff
@@ -654,9 +657,9 @@ context:
 > ![Tip](images/tips.png) _*Notes*_
 > Controlling a central boiler using software or hardware such as home automation can pose risks to its proper functioning. Before using these functions, make sure that your boiler has safety functions and that they are working. Turning on a boiler if all the taps are closed can generate excess pressure, for example.
 
-## Parameters synthesis
+## Parameter summary
 
-| Paramètre                                 | Libellé                                                                       | "over switch" | "over climate"      | "over valve" | "central configuration" |
+| Parameter                                 | Description                                                                   | "over switch" | "over climate"      | "over valve" | "central configuration" |
```
pdcastro, line 662: Oh, and I have also amended this table since your last review.

README-fr.md
```diff
@@ -245,6 +245,9 @@ En fonction de votre choix sur le type de thermostat, vous devrez choisir une ou
 
 ### Pour un thermostat de type ```thermostat_over_switch```
 ![image](images/config-linked-entity.png)
+
+Certains équipements nécessitent d'être périodiquement sollicités pour les empêcher un arrêt de sécurité. Connu sous le nom de "keep-alive" cette fonction est activable en entrant un nombre de secondes non nul dans le champ d'intervalle keep-alive du thermostat. Pour désactiver la fonction ou en cas de doute, laissez-le vide ou entrez zéro (valeur par défaut).
```
jmcollin78: "Certains équipements nécessitent d'être périodiquement sollicités pour  empêcher un arrêt de sécurité": remove 'les'

custom_components/versatile_thermostat/translations/en.json
jmcollin78: Great.

custom_components/versatile_thermostat/translations/fr.json
```diff
@@ -67,6 +68,7 @@
           "heater_entity2_id": "Optionnel entity id du 2ème radiateur",
           "heater_entity3_id": "Optionnel entity id du 3ème radiateur",
           "heater_entity4_id": "Optionnel entity id du 4ème radiateur",
+          "heater_keep_alive": "Intervalle de rafraichissement du switch en secondes. Mettre 0 ou vide pour désactiver. À n'utiliser que pour les switchs qui le nécessite.",
```
jmcollin78: "Intervalle de rafraichissement du switch en secondes. Laisser vide pour désactiver. À n'utiliser que pour les switchs qui le nécessite."

custom_components/versatile_thermostat/translations/fr.json
```diff
@@ -314,6 +317,7 @@
...
```
jmcollin78: idem

pdcastro:
I think it’s all done! 🎉

jmcollin78 (approved):
Nice work !

jmcollin78:
Merged !

PR #350, opened 2024-01-19
Change log level of “Window auto event is ignored” log message

pdcastro:
Since updating to a recent version of Versatile Thermostat (I think v5.2.1 or later), my Home Assistant logs are flooded with the following info log messages (printed every 30 seconds), _even when the “Use window detection” checkbox is **unticked** in the configuration:_

```
2024-01-19 14:41:36.655 INFO (MainThread) [custom_components.versatile_thermostat.base_thermostat] ... - Window auto event is ignored because bypass is ON or window auto detection is disabled
2024-01-19 14:42:06.693 INFO (MainThread) [custom_components.versatile_thermostat.base_thermostat] ... - Window auto event is ignored because bypass is ON or window auto detection is disabled
2024-01-19 14:42:36.613 INFO (MainThread) [custom_components.versatile_thermostat.base_thermostat] ... - Window auto event is ignored because bypass is ON or window auto detection is disabled
2024-01-19 14:43:06.631 INFO (MainThread) [custom_components.versatile_thermostat.base_thermostat] ... - Window auto event is ignored because bypass is ON or window auto detection is disabled
...
```

I reckon that this log message is not needed when the window detection feature is not enabled by the user (my case). Interestingly, I also observed that if the “Use window detection” checkbox was ***ticked,*** the message above was not printed (or at least it was not printed every 30 seconds).

This PR proposes a fix to this issue.

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -1859,7 +1859,7 @@ class BaseThermostat(ClimateEntity, RestoreEntity):
         )
 
         if self.window_bypass_state or not self.is_window_auto_enabled:
-            _LOGGER.info(
+            _LOGGER.debug(
                 "%s - Window auto event is ignored because bypass is ON or window auto detection is disabled",
                 self,
             )


jmcollin78 (changes requested):
Many thanks for this PR,

This is not what we should do. If we do that I the slope will not be calculated at all.
I know the slope calculation is done in a window part for historical reason.
I'm pretty sure tests don't work with this code. Are they  ?

The better thing to do is to extract the slope calculation from the window algo and put it in the "normal" loop (ie in the async_control_heating). 

But I guess this is something tricky and that should break some tests.

For this PR, you could:
- add the _use_window_feature you propose in the post_init to avoid calculating it each timen
- add a is_window_enabled @property which returns the _use_window_feature
- change the log level to .debug to remove the recurrent log.

pdcastro:
> This is not what we should do. If we do that I the slope will not be calculated at all. I know the slope calculation is done in a window part for historical reason. I'm pretty sure tests don't work with this code. Are they ?

I see, I hadn’t noticed the side effects on the slope calculation, sorry! (The tests did pass.)

> For this PR, you could:
> [...] change the log level to .debug to remove the recurrent log.

So, would you be happy with the log level being changed from info to debug? My core motivation for this PR was preventing that `info` log message being repeatedly printed while I was not even using the window feature. :-) I could edit the title of this PR to something like _“Amend log level for window auto event log message“_ and make it a single-line PR changing just the log level.

Thanks for the review. 👍

jmcollin78:
Sorry you have conflict now, because I have done much work this week-end. What you can do, is to merge or close this PR and open a new one, just with the log level change. I will merge it rapidly to avoid more conflict.

jmcollin78 (approved):
Thank you for this PR !

PR #351, opened 2024-01-19
Only modify _last_regulation_change if new regulation is sent to thermostats

ikoz:
This MR fixes and issue where `_last_regulation_change` is set to `now` even though the new temperature setting is never sent to the thermostats, because `abs(dtemp) < self._auto_regulation_dtemp`


jmcollin78 (approved):
Seems good indeed. Thank you for this.

PR #364, opened 2024-01-25
Add more type hints in the thermostat classes and selected files

pdcastro:
This PR adds more type hints to selected files, mainly the thermostats classes. It started with wanting to know the type of the `config_entry` and `entry_infos` variables in the code, as I wondered if it was an object that had the same `get(key[, default])` method signature as the standard `dict` class. I found that yes, it does. Then I thought, _“Why don’t I add the type hint then?”_ From there it snowballed a bit and became this PR, but I think I am stopping here (I am not planning further PRs related to typing).

I don’t see type hints as just “hints for the developer that are ignored by the Python interpreter”, I see them as enabling the use of type checking tools such as `pyright` that help to avoid and catch bugs. Running `pyright` in the dev container (installed with `pip install pyright`) reported an _increase_ from 379 to 412 errors before and after this PR respectively. The increased number errors is a good thing in the sense that the additional errors were simply being masked by the lack of type information. Pyright does not report an error if anything is assigned to a variable of type `Unknown` or `Any`, whereas it is an error to assign, say, a `str` value to a variable of type `float | None` for example.

Just glancing over the errors reported by pyright, I found at least one that points to code that cannot be right:

```
thermostat_climate.py:785:26 - error: Cannot access member "hvac_modes" for type "type[super]”
```

It is a line that reads `super.hvac_modes` instead of `super().hvac_modes`.

However, to confirm, this PR does **not** make any changes to functionality, nor fixes any issues. It just adds type hints. All tests pass with `pytest` and the code was formatted with `black` in the dev container.

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -6,6 +6,8 @@ import math
 import logging
 
 from datetime import timedelta, datetime
+from types import MappingProxyType
+from typing import Any
 
 from homeassistant.util import dt as dt_util
 from homeassistant.core import (
@@ -20,10 +22,12 @@ from homeassistant.components.climate import ClimateEntity
 from homeassistant.helpers.restore_state import RestoreEntity
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.helpers.device_registry import DeviceInfo, DeviceEntryType
+from homeassistant.helpers.typing import EventType as HASSEventType
 
 from homeassistant.helpers.event import (
     async_track_state_change_event,
     async_call_later,
+    EventStateChangedData,
 )
 
 from homeassistant.exceptions import ConditionError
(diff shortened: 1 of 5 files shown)


custom_components/versatile_thermostat/base_thermostat.py
```diff
@@ -134,6 +138,7 @@
 from .ema import ExponentialMovingAverage
 
 _LOGGER = logging.getLogger(__name__)
+ConfigData = Mapping[str, Any]
```
pdcastro: Correction: This type should be `MappingProxyType`, not `Mapping`. More details in [another comment](https://github.com/jmcollin78/versatile_thermostat/pull/345#discussion_r1465733083). I am force-pushing this change.

jmcollin78:
Pushing to github pages failed but this is useless so I will merge.

PR #367, opened 2024-01-27
README*.md: Replace absolute image URLs with relative URLs

pdcastro:
While adding documentation for PR #345, I edited configuration image files on my local workstation for the README files and previewed them locally with relative URLs, but I noticed that the majority of existing image URLs are absolute.

Is there a reason to prefer absolute image URLs in the README files? Relative URLs have the advantage that it is easier to preview modifications locally before committing the files to Github.

As for the `?raw=true` query string at the end of the absolute URLs, I gather that it causes a HTTP 302 redirect:

```sh
$ curl -v https://github.com/jmcollin78/versatile_thermostat/blob/main/images/new-icon.png?raw=true >/dev/null
...
< HTTP/2 302
< location: https://github.com/jmcollin78/versatile_thermostat/raw/main/images/new-icon.png
```

```
$ curl -v https://github.com/jmcollin78/versatile_thermostat/blob/main/images/new-icon.png >/dev/null
...
< HTTP/2 200
```

I understand that this redirect is useful when linking files from outside of Github, but in the README files (“inside Github”) I think it just slows things down a bit.

As an experiment, this PR replaces all absolute URLs with relative URLs, and deletes the `?raw=true` query string. I think this improves performance both locally on VSCode (the image files are loaded locally) and on Github (no redirects).

I have previewed the changes on Github by clicking on the three horizontal dots to the right of the file names and selecting “View file”, and I have previewed the changes locally on VS Code. It all looks fine. I cannot guarantee that there aren’t side effects, but if there are, it would also be straightforward to revert the changes with either `git checkout <commit> -- README.md` or by reinserting the URLs with “find and replace” in VS Code.

### DIFF
--- a/README-fr.md
+++ b/README-fr.md
@@ -4,9 +4,9 @@
 [![hacs][hacs_badge]][hacs]
 [![BuyMeCoffee][buymecoffeebadge]][buymecoffee]
 
-![Tip](https://github.com/jmcollin78/versatile_thermostat/blob/main/images/icon.png?raw=true)
+![Tip](images/icon.png)
 
-> ![Tip](https://github.com/jmcollin78/versatile_thermostat/blob/main/images/tips.png?raw=true) Cette intégration de thermostat vise à simplifier considérablement vos automatisations autour de la gestion du chauffage. Parce que tous les événements autour du chauffage classiques sont gérés nativement par le thermostat (personne à la maison ?, activité détectée dans une pièce ?, fenêtre ouverte ?, délestage de courant ?), vous n'avez pas à vous encombrer de scripts et d'automatismes compliqués pour gérer vos climats. ;-).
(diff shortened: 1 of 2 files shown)


jmcollin78:
Hello, yes I have started let's say a POC for this and I don't have generalization yet. This is what you have done ! Fine.
A simple question why relative path `images/xxxx` and not absolute `/images/xxx` which seems to work well also ?

I begin with the absolute path.

jmcollin78:
Hello @pdcastro , thank you for this. I have started, let's say a POC, to replace some URLs. You have seen it.

You have to know that I do some copy (markdown copy) to some forums (see https://forum.hacf.fr/t/nouveau-thermostat-type-proportionnel-avec-gestion-des-presets-portes-et-fenetres-detection-de-mouvement-gestion-de-presence-et-surconsommation/19160) and I guess, relative images won't work or I will have to replace all relative path with absolute path pointing the github images repository. That's why the URLs was absolute into the Github main images dir.

Same explaination for ?raw=true.

But the drawback is that when editing the README in local, you don't see the new images, I agree this is annoying.
I will give it a try, nothing difficult to reverse if this make some weird effects.

jmcollin78 (approved):

jmcollin78:
I test on the forum copying the README-fr and it breaks all images:
Capture d’écran 2024-01-27 à 09 45 06

So I test and I need the absolute github path AND the ?raw=true else if don't work.
So I will to do an intelligent copy/paste of the markdown file.

Let's keep it like that. Thank you once again.

PR #374, opened 2024-02-04
Prevent “Heating will be disabled” warning loop while HVACMode is OFF

pdcastro:
My Home Assistant logs were full of the following warning messages (printed about every 30 seconds) even though Versatile Thermostat was turned off (`self._hvac_mode is HVACMode.OFF`):

```
2024-02-04 17:02:53.951 WARNING (MainThread) [custom_components.versatile_thermostat.prop_algorithm] Proportional algorithm: calculation is not possible cause target_temp or current_temp is null. Heating/cooling will be disabled
2024-02-04 17:03:23.946 WARNING (MainThread) [custom_components.versatile_thermostat.prop_algorithm] Proportional algorithm: calculation is not possible cause target_temp or current_temp is null. Heating/cooling will be disabled
2024-02-04 17:03:53.978 WARNING (MainThread) [custom_components.versatile_thermostat.prop_algorithm] Proportional algorithm: calculation is not possible cause target_temp or current_temp is null. Heating/cooling will be disabled
...
```

`current_temp` was indeed `None`, because I had unplugged a MQTT heater + temperature sensor from the wall socket while it was not being used.

It is a useful warning message while Versatile Thermostat is turned on (`HVACMode.HEAT`, `HVACMode.COOL`), but I think it is counterproductive while Versatile Thermostat is turned off (`HVACMode.OFF`). Warning that _“Heating/cooling will be disabled”_ when the user has already turned heating/cooling off is redundant. The warning that `target_temp` or `current_temp` are null is interesting but not in a loop while `HVACMode` is `OFF`, in a scenario like mine where the controlled heater + temperature sensors were unplugged.

This PR proposes dynamically amending the log level of that message depending on the on/off state of Versatile Thermostat.

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -799,7 +799,7 @@ class BaseThermostat(ClimateEntity, RestoreEntity):
                         self._target_temp,
                         self._cur_temp,
                         self._cur_ext_temp,
-                        self._hvac_mode == HVACMode.COOL,
+                        self._hvac_mode or HVACMode.OFF,
                     )
 
             self.hass.create_task(self._check_initial_state())
--- a/custom_components/versatile_thermostat/prop_algorithm.py
+++ b/custom_components/versatile_thermostat/prop_algorithm.py
@@ -1,6 +1,8 @@
 """ The TPI calculation module """
 import logging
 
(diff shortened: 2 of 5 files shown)


jmcollin78 (approved):
Wonderful ! Thank you for th!s

PR #382, opened 2024-02-16
Make the switch keep-alive callback conditional on the entity state

pdcastro:
This PR is motivated by discussion #379. It is *not* clear that it fixes any bug, because we have *not* identified a bug: We have not been able to reproduce the issue reported by the user. All my tests — both automated unit tests and manual end-to-end tests — behaved the same way before and after this PR.

Having said that, the added logic in this PR fends off against some theoretical speculation of what might happen in some corner cases. See other comments for more details.

### DIFF
--- a/custom_components/versatile_thermostat/underlyings.py
+++ b/custom_components/versatile_thermostat/underlyings.py
@@ -220,9 +220,7 @@ class UnderlyingSwitch(UnderlyingEntity):
     @overrides
     def startup(self):
         super().startup()
-        self._keep_alive.set_async_action(
-            self.turn_on if self.is_device_active else self.turn_off
-        )
+        self._keep_alive.set_async_action(self._keep_alive_callback)
 
     # @overrides this breaks some unit tests TypeError: object MagicMock can't be used in 'await' expression
     async def set_hvac_mode(self, hvac_mode: HVACMode) -> bool:
@@ -247,9 +245,14 @@ class UnderlyingSwitch(UnderlyingEntity):
             not self.is_inversed and real_state
         )
 
+    async def _keep_alive_callback(self):
+        """Keep alive: Turn on if already turned on, turn off if already turned off."""
+        await (self.turn_on() if self.is_device_active else self.turn_off())
+
     # @overrides this breaks some unit tests TypeError: object MagicMock can't be used in 'await' expression
     async def turn_off(self):
         """Turn heater toggleable device off."""
+        self._keep_alive.cancel()  # Cancel early to avoid a turn_on/turn_off race condition
         _LOGGER.debug("%s - Stopping underlying entity %s", self, self._entity_id)
         command = SERVICE_TURN_OFF if not self.is_inversed else SERVICE_TURN_ON
         domain = self._entity_id.split(".")[0]
@@ -258,7 +261,7 @@ class UnderlyingSwitch(UnderlyingEntity):
             try:
                 data = {ATTR_ENTITY_ID: self._entity_id}
                 await self._hass.services.async_call(domain, command, data)
-                self._keep_alive.set_async_action(self.turn_off)
+                self._keep_alive.set_async_action(self._keep_alive_callback)
             except Exception:
                 self._keep_alive.cancel()
                 raise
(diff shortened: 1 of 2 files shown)


custom_components/versatile_thermostat/underlyings.py
```diff
@@ -247,9 +245,14 @@ def is_device_active(self):
             not self.is_inversed and real_state
         )
 
+    async def _keep_alive_callback(self):
+        """Keep alive: Turn on if already turned on, turn off if already turned off."""
+        await (self.turn_on() if self.is_device_active else self.turn_off())
```
pdcastro, line 250: Some comments in #379 suggested / implied that the keep-alive logic was responsible for incorrectly turning on a switch that should have been turned off. I don't see how that is possible because if the switch was turned off, then the `turn_off` should have been successfully called, and if the `turn_off` method was successfully called, then any `turn_on` keep-alive callback would have been replaced with a `turn_off` callback that keeps the switch turned off.

Nonetheless, I also note that some actions like setting the climate preset to "Off" only _conditionally_ call the `turn_off` method, for example:

```python
    async def set_hvac_mode(self, hvac_mode: HVACMode) -> bool:
        if hvac_mode == HVACMode.OFF:
            if self.is_device_active:
                await self.turn_off()
```
https://github.com/jmcollin78/versatile_thermostat/blob/5.4.1/custom_components/versatile_thermostat/underlyings.py#L232-L233

I think the overall logic is too complex for us to be able to readily “prove” that the `turn_off` method is always called before or after the preset mode is set to "Off", and if the `turn_off` method was *not* called, while the `turn_on` method was previously called, then the keep-alive callback would keep calling the `turn_on` method.

The logic added above is a direct reaction to this hypothetical reasoning.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -247,9 +245,14 @@ def is_device_active(self):
...
+
     # @overrides this breaks some unit tests TypeError: object MagicMock can't be used in 'await' expression
     async def turn_off(self):
         """Turn heater toggleable device off."""
+        self._keep_alive.cancel()  # Cancel early to avoid a turn_on/turn_off race condition
```
pdcastro, line 255: This `cancel()` call here and similarly in the `turn_on()` method is something that I had added to my original keep-alive PR (#345), and you pushed back saying that it was _“Not necessary because `set_async_action` already cancels”_ ([link](https://github.com/jmcollin78/versatile_thermostat/pull/345#discussion_r1459450453)). That‘s true, but I had pointed out ([link](https://github.com/jmcollin78/versatile_thermostat/pull/345#discussion_r1460058632)) that there is an async call (`await` statement) between this `cancel()` call and the `set_async_action()` call, which causes an async context switch during which other queued async tasks get a chance to run. An imminent keep-alive callback could be among those other queued tasks, and theoretically a `turn_on` callback could be called in the middle of a `turn_off` method. In the original PR, I had ultimately reasoned that even if this happened, the correct service call should immediately follow, so I treated this early `cancel()` call as an optimisation that was not sufficiently justified and thus I agreed to delete it.

However, I cannot “prove” that preventing a `turn_on` callback from being called in the middle of the `turn_off` method is ”just an optimisation” rather than a ”correctness requirement,” and therefore I prefer to make this early `cancel()` call to simply avoid this situation in the first place — especially in the context of suspicions (#379). If you can prove that the early `cancel()` call is not required, then we can delete it again. 🧐😀

Having said that, the user (#379) claimed that their issue was (consistently?) reproducible and my reasoning above should only apply to ”rare corner cases.” Still, I’d rather fend off against rare corner cases as well.

tests/test_switch_keep_alive.py
```diff
@@ -210,6 +210,7 @@ async def test_switch_keep_alive(self, common_mocks: CommonMocks):
             common_mocks,
             [call("switch", SERVICE_TURN_ON, {"entity_id": "switch.mock_switch"})],
         )
+        common_mocks.mock_is_state.return_value = True
```
pdcastro, line 213: Because the callback calls are now conditional on `self.is_device_active`, and the Home Assistant state is mocked.

jmcollin78 (approved):
Nice ! Thank you for your work.

PR #417, opened 2024-03-20
English translations

EPicLURcher:
-Translated strings in en.json
-Copied entire contents of en.json into strings.json


jmcollin78 (approved):
Thank you !

PR #453, opened 2024-04-23
Enhance temperature regulation when working with internal device temperature

jdeksup:
Hi,
I use Versatile thermostat from several weeks and really appreciate it to regulate my climate devices that have a bad thermal regulation because of the position of the sensor. Thank you for your big work on this. 
I suggest a small adjustment : a behavior that prevents the system from forgetting the regulation when using the device's internal temperature as an offset.

Although the regulation itself remains unchanged, variations in the device's internal temperature necessitate adjustments to the target temperature to maintain effective regulation. This code modification specifically bypasses the forget mechanism for minor adjustments, while the forget mechanism based on duration is still maintained.

Here is an history of the underlying climate before : 
image
and after (with 2 minutes of interval which is fast) : 
image

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -168,11 +168,17 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
 
         _LOGGER.info("%s - regulation calculation will be done", self)
 
+        # use _attr_target_temperature_step to round value if _auto_regulation_dtemp is equal to 0
+        regulation_step = self._auto_regulation_dtemp if self._auto_regulation_dtemp else self._attr_target_temperature_step
+        _LOGGER.debug("%s - usage of regulation_step: %.2f ",
+                      self,
+                      regulation_step)
+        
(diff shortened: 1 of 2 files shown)


jmcollin78:
Hello @jdeksup ,

I understand the idea and it could be interesting. It has one important drawback: we should avoid to send to much regulation command (setpoint change) on some equipment like TRV. Else we will drawn the battery very fast. (see some issue relative to this particular point).

This threshold have been done to avoid this battery drawn and I fear that your improvement will rollback this feature.

jmcollin78:
> I understand your point. I already tried to configure dtemp threshold to 0 and got an error "assert x > 0" on round_to_nearest method, So I just update my pull request to fix this and it works.

This seems good. Can you please add a unit test with dtemp equal to 0 to prove the correct behavior ?
You have a good example in the test method:
```
async def test_over_climate_regulation(
    hass: HomeAssistant, skip_hass_states_is_state, skip_send_event
):
```

in the `tests/test_auto_regulation.py`. Of course I can help if needed.

jmcollin78 (approved):
Thank you  @jdeksup

Issue #240, opened 2023-11-30
Fahrenheit conversion is not correct, seems to make the conversion twice. 


darkwolfgar:
When converting Fahrenheit temperatures to Celsius and putting the Celsius temperature in the configuration settings, this seems to be correct, See below. 
![image](https://github.com/jmcollin78/versatile_thermostat/assets/88863932/d842defb-386b-46e3-975e-a0ffe4ad0abf)
However, as shown in the picture below, the current temperature is taking the Fahrenheit temperature of my sensor and using the Fahrenheit to Celsius conversion on it again.
![image](https://github.com/jmcollin78/versatile_thermostat/assets/88863932/bb6bab7f-516f-4605-b53c-d99461ec4925)

darkwolfgar:
Also it appears that even though its showing Fahrenheit, when I set the temperature to 75, on my thermostat its setting it to 45 degrees Fahrenheit instead of 75.

jmcollin78:
It seems you are in `over_climate`. It that mode it should use the temperature unit of the underlying climate. Is the underlying climate correctly displayed ?

jmcollin78:
I never do any translation of °C to °F or reverse. I let the system do like it always does.

The unit of all VTherm is the unit configured in the HA configuration.
`self._unit = self._hass.config.units.temperature_unit` 

Is your global HA configuration set the Kelvin ?
Is your underlying climate also configured in °K ? You should be able to see that in the attributes of the climate (Dev Tools / States).

jmcollin78:
When I switch my test env with the checkox checked, I have this :

![Capture d’écran 2023-12-03 à 19 37 41](https://github.com/jmcollin78/versatile_thermostat/assets/1712374/e43e9261-bdae-41c7-8e4b-ff326e6b5987)


This seems coherent. I guess the problem is not related to VTHerm itself.

jmcollin78:
How to you set the temperature for preset in the VTherm configuration ? Do you set the temperature in °C or °K ?

jmcollin78:
I can't understand why this behavior. Even if I configure my HA into °F I don't have this but my sensor are all in °C.
Can you please post me the attributes of the VTherm (in Development  Tools / State, find your VTherm and post the attributes in yaml formatted with the <> button).
I want to check if the unit is correctly set.

jmcollin78:
Hello @benjinne ,

[...]

PR #460, opened 2024-05-12
[Draft] Use user's preferred temperature unit instead of hardcoding celsius

mbbush:
With this change, I'm no longer affected by #240. 

I tried to write a unit test for this behavior, which would fail on main and pass on my branch, but I'm unfamiliar with python testing libraries, or how home assistant uses them, and wasn't able to figure it out.


jmcollin78 (approved):
Maybe you find what goes wrong. Thank you

PR #461, opened 2024-05-14
Change VTherm temperature unit to HA's preferred unit.

mag2352:
With this change, I no longer suffer from #240. 

I made 2 basic changes to the code, simplifying the changes I mentioned in #240. 

- Firstly, I changed the unit used by VTherm to match HA's preferred unit.
- Secondly, I changed the references to the underlying entity's temperatures to use HA's state references. This automatically includes HA's converted values, and should be robust against having mismatched temperature units. I propose that all references to the underlying entity use HA's state, rather than directly accessing the values from the entity.

Example:

Change: 
`max_val = self._underlying_climate.max_temp`

To:
`max_val = self._hass.states.get(self._entity_id).attributes.get("max_temp")`

This results in the correct converted value being grabbed if the underlying entity uses a different unit of temperature than HA's preferred unit. This should make no difference for those who have entities that match their HA's preferred unit. I only made the minimum changes to allow for my use case to work, but there may be other attributes that should be changed to reflect this.

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -207,7 +207,7 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
                 # regulation can use the device_temp
                 self.auto_regulation_use_device_temp
                 # and we have access to the device temp
-                and (device_temp := under.underlying_current_temperature) is not None
+                and (device_temp := self._hass.states.get(under._entity_id).attributes.get("current_temperature")) is not None
                 # and target is not reach (ie we need regulation)
                 and (
                     (
@@ -900,10 +900,7 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
     @property
     def temperature_unit(self) -> str:
         """Return the unit of measurement."""
-        if self.underlying_entity(0):
-            return self.underlying_entity(0).temperature_unit
-
-        return self._unit
+        return self.hass.config.units.temperature_unit
 
     @property
     def supported_features(self):
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -207,7 +207,7 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
                 # regulation can use the device_temp
                 self.auto_regulation_use_device_temp
                 # and we have access to the device temp
-                and (device_temp := self._hass.states.get(under._entity_id).attributes.get("current_temperature")) is not None
+                and (device_temp := under.underlying_current_temperature) is not None
                 # and target is not reach (ie we need regulation)
                 and (
                     (
--- a/custom_components/versatile_thermostat/underlyings.py
+++ b/custom_components/versatile_thermostat/underlyings.py
@@ -30,6 +30,7 @@ from homeassistant.components.number import SERVICE_SET_VALUE
 
 from homeassistant.helpers.entity_component import EntityComponent
 from homeassistant.helpers.event import async_call_later
+from homeassistant.util.unit_conversion import TemperatureConverter
 
 from .const import UnknownEntity, overrides
 from .keep_alive import IntervalCaller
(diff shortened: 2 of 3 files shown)


mbbush:
This approach makes sense to me, although I haven't tested it yet or looked for other places to change.

mag2352:
> This approach makes sense to me, although I haven't tested it yet or looked for other places to change.

I would imagine there are other places that would require changes. Notably, I didn't perform any changes on the temperature sensors. I used the (rather loose) assumption that the temperature sensor being used would match HA's preferred unit, but it is fairly trivial to check the unit of measurement on the sensor, and convert if needed.

jmcollin78:
Hello @mag2352 ,
This PR is still tagged as draft. Do you think I will be able to merge it and do a draft release ?

mag2352:
> Hello @mag2352 , This PR is still tagged as draft. Do you think I will be able to merge it and do a draft release ?

I think the mag2352/versatile_thermostat@f8e57b3eb7beda6cc481211fd4bbb854e8172d87 can be merged without issue (I removed the subsequent commits for now and moved them to another branch as I'm still testing those changes).

custom_components/versatile_thermostat/thermostat_climate.py
```diff
@@ -207,7 +207,7 @@ async def _send_regulated_temperature(self, force=False):
                 # regulation can use the device_temp
                 self.auto_regulation_use_device_temp
                 # and we have access to the device temp
-                and (device_temp := under.underlying_current_temperature) is not None
+                and (device_temp := self._hass.states.get(under._entity_id).attributes.get("current_temperature")) is not None
```
jmcollin78: Why this change ? Isn't it supposed to do exactly the same ?

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -731,8 +731,8 @@ def cap_sent_value(self, value) -> float:
             self._underlying_climate.min_temp is not None
             and self._underlying_climate is not None
         ):
-            min_val = self._underlying_climate.min_temp
-            max_val = self._underlying_climate.max_temp
+            min_val = self._hass.states.get(self._entity_id).attributes.get("min_temp")
+            max_val = self._hass.states.get(self._entity_id).attributes.get("max_temp")
```
jmcollin78: Why this change ? Is that not exactly the same result ?

mag2352: It is not exactly the same result. Your original method uses the device's raw states. Using the state from HA uses the auto-converted states from HA. A user does not have the option to set what unit a climate entity is in, so accessing the underlying climate entity's temperatures and temperature ranges will return these values in their raw states (in Celsius even if HA is set to Fahrenheit). We don't want this, since this results in double conversion, and inaccurate values for the cap function you have. This would make no difference for someone that has a climate entity in Celsius and their default unit in Celsius, but fixes the problem for mismatched climate entity units.

Same justification applies to your other comment. I am now seeing it might be better to directly change the properties themselves in underlying.py, but the end result is the same. The whole idea is that there is a difference between raw values and values that HA use. So even if an entity has a unit of F due to HA's auto conversion, its raw state (which you are accessing in the original code) is not actually in that unit.

State vs Raw State (This also applies to climate entities, just showing a sensor for simplicity): 
image

jmcollin78: ok. Thank you for the explaination.
I will merge this PR and create a release.

Many thanks for your time to solve this old issue.

jmcollin78 (dismissed):

jmcollin78:
Unit tests are failing. Can you please have a look of what happens ?
The error seems not clear for me.

jmcollin78 (approved):

jmcollin78:
unit tests still failed on dependencies

mag2352:
Yeah, I'm probably going to test this locally for now. I'm assuming its a pytest error, as that's what seems to be the most likely problem based on the first error message.

jmcollin78:
Why is there any dependencies changes for so minor changes ?

mag2352:
I'm not sure what is causing it, this is definitely strange. Wondering if some dependency updated recently and is causing this issue.

mag2352:
Passes Unit Tests on my repository. Had to change the cap_sent_value function to explicitly convert temperature values if needed rather than accessing HA's state (since it would sometimes return none when getting the entity). I have never had an issue with this when running on HA (I have been running my branch for weeks now without issue), but this is functionally the same and passes your unit tests.

PR #471, opened 2024-06-15
issue with ac and mouvement detection

cddu33:
https://github.com/jmcollin78/versatile_thermostat/issues/470

issue with ac and mouvement detection


jmcollin78 (commented):
Thank you for this fix. There is already a method named find_preset_temp which calculate the preset_name depending on the AC mode (and other stuff).

Can we use it in that case ?

cddu33:
No because motion presept not use it.

cddu33:
i modify function find_preset_temps for use with activity preset
j'ai modifier la fonction find_preset_temps pour être utilisable avec le module Activité (détection de mouvement)

jmcollin78 (approved):
Thank you ! Will check if unit tests are ok. Not totally sure it will pass as is.

jmcollin78:
All is fine. I will release soon.

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/releases/tag/6.2.5

Issue #508, opened 2024-09-27
errors on new HA beta


tbrasser:
Logger: homeassistant.core
Bron: core.py:2781
Eerst voorgekomen: 17:03:58 (25 gebeurtenissen)
Laatst gelogd: 18:07:53

Error executing service: <ServiceCall climate.set_temperature (c:01J8T3SZ09TWNKCWKY6S4CV274): entity_id=['climate.trv_keuken'], temperature=19.2, target_temp_high=25.0, target_temp_low=15.0>
Error executing service: <ServiceCall climate.set_temperature (c:01J8T3SZ0B2B9QGKH93B2NG70J): entity_id=['climate.trv_keuken'], temperature=19.2, target_temp_high=25.0, target_temp_low=15.0>
Error executing service: <ServiceCall climate.set_temperature (c:01J8T3SZ0C2BF9YRWY56BG62QS): entity_id=['climate.trv_badkamer'], temperature=14.0, target_temp_high=25.0, target_temp_low=15.0>
Error executing service: <ServiceCall climate.set_temperature (c:01J8T3T6YX6V0R7YXGWWHW5W66): entity_id=['climate.trv_eetkamer_links'], temperature=23.2, target_temp_high=25.0, target_temp_low=15.0>
Error executing service: <ServiceCall climate.set_temperature (c:01J8T3T6Z14E6JKF3G1NQ7V0PA): entity_id=['climate.trv_eetkamer_rechts'], temperature=23.6, target_temp_high=25.0, target_temp_low=15.0>
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/core.py", line 2781, in _run_service_call_catch_exceptions
    await coro_or_task
  File "/usr/src/homeassistant/homeassistant/core.py", line 2804, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 996, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1068, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/climate/__init__.py", line 977, in async_service_temperature_set
    raise ServiceValidationError(
[...]

PR #509, opened 2024-09-27
Update const.py with SUPPORT_FLAGS ClimateEntityFeature.TARGET_TEMPERATURE_RANGE

tbrasser:
as per #508

### DIFF
--- a/custom_components/versatile_thermostat/const.py
+++ b/custom_components/versatile_thermostat/const.py
@@ -330,7 +330,7 @@ CONF_WINDOW_ACTIONS = [
     CONF_WINDOW_ECO_TEMP,
 ]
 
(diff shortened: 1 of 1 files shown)


tbrasser:
Hmm, no this obviously make my heating only vtherms over climate look like they can heat/cool.

jmcollin78:
> Hmm, no this obviously make my heating only vtherms over climate look like they can heat/cool.

I don't understand this comment. What does it mean exactly ?

tbrasser:
My vtherms show up as heating & cooling devices. While they can only heat.

I think the fix is like:

- read from underlying the supported features
- pass only temperature or also the min/max depending on whether the underlying supports it.

PR #551, opened 2024-10-13
Fix #539 - allow up to 9 entities in over_climate, over_switch and over_valve config forms

### LINKED ISSUE #539: It seems I can't add a fifth heater for an over switch VTherm
I want to add a 5th heater commuter for a single VTherm, how can I do that ?

image

As you can see, the choices are blocked to only 4 inputs.

pounard:
Following your advice in https://github.com/jmcollin78/versatile_thermostat/issues/539#issuecomment-2408473996, I tried!

OK so you have to know that Python is not the language I work with in a daily basis...

Some notes:
 - Adding the item into the UI is quite trivial.
 - Then patching whenever you used that value is also, because your code is well designed, and simply create some list and use that afterwards, no matter how much entities there are.

But, some caveat, it seems to alter some calculations:
 - In `tests/test_multiple_switch.py` there are a few `call.call_later(hass, 120.0, ANY)` which I assume assert that some events are triggered later, but what I suspect to be timings do change. I have no idea how this works and why, but changing the values make the test happy.
 - In `tests/test_multiple_switch.py` there is a calculation I do not understand which states in a comment `100 of the device / 5 -> 20, current power 50 so max is 75` so I had to change a few values there by trial and error until I got it right, I guess it need a throughout review.

Please note I did full TDD here, I didn't test it in real conditions, but I guess it should work.

I'm not accustomed enough with your test suite and scripts, and I have no idea how to run some test hassio instance with the dev code within. Any help would be greatly appreciated.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue.md
+++ b/.github/ISSUE_TEMPLATE/issue.md
@@ -94,6 +94,8 @@ underlying_switch_0: input_boolean.fake_heater_switch1
 underlying_switch_1: null
 underlying_switch_2: null
 underlying_switch_3: null
+underlying_switch_4: null
+underlying_switch_5: null
 on_percent: 0.1
 on_time_sec: 6
 off_time_sec: 54
--- a/tests/test_central_boiler.py
+++ b/tests/test_central_boiler.py
@@ -251,6 +251,7 @@ async def test_update_central_boiler_state_multiple(
     switch2 = MockSwitch(hass, "switch2", "theSwitch2")
     switch3 = MockSwitch(hass, "switch3", "theSwitch3")
     switch4 = MockSwitch(hass, "switch4", "theSwitch4")
+    switch5 = MockSwitch(hass, "switch5", "theSwitch5")
 
     entry = MockConfigEntry(
         domain=DOMAIN,
@@ -276,6 +277,7 @@ async def test_update_central_boiler_state_multiple(
             CONF_HEATER_2: switch2.entity_id,
             CONF_HEATER_3: switch3.entity_id,
             CONF_HEATER_4: switch4.entity_id,
+            CONF_HEATER_5: switch5.entity_id,
             CONF_PROP_FUNCTION: PROPORTIONAL_FUNCTION_TPI,
             CONF_INVERSE_SWITCH: False,
             CONF_TPI_COEF_INT: 0.3,
@@ -302,6 +304,7 @@ async def test_update_central_boiler_state_multiple(
     assert entity.underlying_entities[1].entity_id == "switch.switch2"
     assert entity.underlying_entities[2].entity_id == "switch.switch3"
     assert entity.underlying_entities[3].entity_id == "switch.switch4"
+    assert entity.underlying_entities[4].entity_id == "switch.switch5"
 
     assert api.nb_active_device_for_boiler_threshold == 1
     assert api.nb_active_device_for_boiler == 0
@@ -463,7 +466,37 @@ async def test_update_central_boiler_state_multiple(
         assert api.nb_active_device_for_boiler == 4
         assert boiler_binary_sensor.state == STATE_ON
 
-    # 5. stop a heater
+    # 5. start a 5th heater
+    with patch(
+        "homeassistant.core.ServiceRegistry.async_call"
+    ) as mock_service_call, patch(
+        "custom_components.versatile_thermostat.binary_sensor.send_vtherm_event"
+    ) as mock_send_event:
+        await switch5.async_turn_on()
+        switch5.async_write_ha_state()
+        # Wait for state event propagation
+        await asyncio.sleep(0.1)
+
+        assert entity.hvac_action == HVACAction.HEATING
+
+        # Only the first heater is started by the algo
+        assert mock_service_call.call_count == 1
+        # No switch of the boiler. Caution: each time a underlying heater state change itself,
+        # the cycle restarts. So it is always the first heater that is started
+        mock_service_call.assert_has_calls(
+            [
+                call.service_call(
+                    "switch",
+                    "turn_on",
+                    {"entity_id": "switch.switch1"},
+                ),
+            ]
+        )
+        assert mock_send_event.call_count == 0
+        assert api.nb_active_device_for_boiler == 5
+        assert boiler_binary_sensor.state == STATE_ON
+
+    # 6. stop a heater
     with patch(
         "homeassistant.core.ServiceRegistry.async_call"
     ) as mock_service_call, patch(
@@ -478,10 +511,10 @@ async def test_update_central_boiler_state_multiple(
 
         assert mock_service_call.call_count == 0
         assert mock_send_event.call_count == 0
-        assert api.nb_active_device_for_boiler == 3
+        assert api.nb_active_device_for_boiler == 4
         assert boiler_binary_sensor.state == STATE_ON
 
-    # 6. stop a 2nd heater
+    # 7. stop a 2nd heater
     with patch(
         "homeassistant.core.ServiceRegistry.async_call"
     ) as mock_service_call, patch(
@@ -489,6 +522,24 @@ async def test_update_central_boiler_state_multiple(
     ) as mock_send_event:
         await switch4.async_turn_off()
         switch4.async_write_ha_state()
+        # Wait for state event propagation
+        await asyncio.sleep(0.1)
+
+        assert entity.hvac_action == HVACAction.HEATING
+
+        assert mock_service_call.call_count == 0
+        assert mock_send_event.call_count == 0
+        assert api.nb_active_device_for_boiler == 3
+        assert boiler_binary_sensor.state == STATE_ON
+
+    # 8. stop a 3rd heater
+    with patch(
+        "homeassistant.core.ServiceRegistry.async_call"
+    ) as mock_service_call, patch(
+        "custom_components.versatile_thermostat.binary_sensor.send_vtherm_event"
+    ) as mock_send_event:
+        await switch5.async_turn_off()
+        switch5.async_write_ha_state()
         # Wait for state event propagation
         await asyncio.sleep(0.1)
 
--- a/tests/test_multiple_switch.py
+++ b/tests/test_multiple_switch.py
@@ -260,6 +260,7 @@ async def test_multiple_switchs(
             CONF_HEATER_2: "switch.mock_switch2",
             CONF_HEATER_3: "switch.mock_switch3",
             CONF_HEATER_4: "switch.mock_switch4",
+            CONF_HEATER_5: "switch.mock_switch5",
             CONF_HEATER_KEEP_ALIVE: 0,
             CONF_MINIMAL_ACTIVATION_DELAY: 30,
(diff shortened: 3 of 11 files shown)


jmcollin78:
Heyy, thank you for this. I will have a look.

I have some major strategic questions before merging:
1. I wonder when this will stop ... if someone wants 6 heaters or 7 heaters. My question what should be the limit for one VTherm. I though was enough. The other users have split into 2 heating zone, which one VTherm for each zone having less than 4 switch. Is that not preferable in fact ?
2. the other question is why only for VTherm `over_switch` ? Why not for `over_valve` and `over_climate` ?

pounard:
> Heyy, thank you for this. I will have a look.
> 
> I have some major strategic questions before merging:
> 
>     1. I wonder when this will stop ... if someone wants 6 heaters or 7 heaters. My question what should be the limit for one VTherm. I though was enough. The other users have split into 2 heating zone, which one VTherm for each zone having less than 4 switch. Is that not preferable in fact ?
> 
>     2. the other question is why only for VTherm `over_switch` ? Why not for `over_valve` and `over_climate` ?

Believe me or not, but I asked myself the exact same questions :sweat_smile: 

Fact is I focused on my own use case, to see whether or not this was doable prior doing anything else.

I guess with some for loops at the right places, the max number of heaters could simply be generated, but it will make all the existing tests obsolete, and that's going to be really hard to fix all those.

pounard:
I wanted to add that, my point of is, that all your tests are very thoroughly built but they are mostly functional tests and not unit tests enough.

I guess that, if I read your code correctly, whatever works for 4 will work for 5, and 6, etc... Simply put, I think coverage can omit building-up 6, 7... entities each time, and focus on a single or two more "unity" tests for testing that configuration is correctly propagated.

pounard:
> I have some major strategic questions before merging:

Please don't merge as-èis, I left a todo about a unit test I messed up but I don't understand what is its goal.

pounard:
I will do a second pass in order to add more than 5 (maybe arbitrarily 10?) and also add them for over_valve and over_climate, is that OK for you ?

pounard:
And did it:
- still untested yet in real conditions,
- there are still 2 places where I'm not sure of what I am doing, please see upcoming commends in diff.

tests/test_multiple_switch.py
```diff
@@ -796,8 +814,9 @@ async def test_multiple_switch_power_management(
     ) as mock_heater_on, patch(
         "custom_components.versatile_thermostat.underlyings.UnderlyingSwitch.turn_off"
     ) as mock_heater_off:
-        # 100 of the device / 4 -> 25, current power 50 so max is 75
-        await send_max_power_change_event(entity, 74, datetime.now())
+        # 100 of the device / 5 -> 20, current power 50 so max is 75
+        # @todo I do not understand this calculation, I arbirarily set "55" instead of the original "74" to make it pass
```
pounard, line 809: Here I didn't understand what your comment meant, so I basically did trial/error until I got the right values... They seemed coherent with the underlaying algorithm in code though.

.github/ISSUE_TEMPLATE/issue.md
```diff
@@ -94,6 +94,11 @@ underlying_switch_0: input_boolean.fake_heater_switch1
 underlying_switch_1: null
 underlying_switch_2: null
 underlying_switch_3: null
+underlying_switch_4: null
+underlying_switch_5: null
+underlying_switch_6: null
+underlying_switch_7: null
+underlying_switch_8: null
```
pounard: I'm not sure about the purpose of those values?

tests/test_central_boiler.py
```diff
@@ -251,6 +251,7 @@ async def test_update_central_boiler_state_multiple(
     switch2 = MockSwitch(hass, "switch2", "theSwitch2")
     switch3 = MockSwitch(hass, "switch3", "theSwitch3")
     switch4 = MockSwitch(hass, "switch4", "theSwitch4")
+    switch5 = MockSwitch(hass, "switch5", "theSwitch5")
```
pounard, line 254: Here I didn't test until 9 heaters because it was a lot of copy/paste/fix of code, maybe the up to 9 limit should be tested in a simpler test ?

tests/test_central_boiler.py
```diff
@@ -276,6 +277,7 @@ async def test_update_central_boiler_state_multiple(
             CONF_HEATER_2: switch2.entity_id,
             CONF_HEATER_3: switch3.entity_id,
             CONF_HEATER_4: switch4.entity_id,
+            CONF_HEATER_5: switch5.entity_id,
```
pounard, line 280: Same as upper.

jmcollin78:
> I will do a second pass in order to add more than 5 (maybe arbitrarily 10?) and also add them for over_valve and over_climate, is that OK for you ?

Sorry but I don't think it is a good thing to multiply the number of underlying with hard coding the limit.
In two years, you are the first that need more than 4 underlyings heaters, so there is not a real and massive need of so much entries. My point was not to say: let's have 10 entries but more, we need a limit and respect this limit. The fallback is very easy: just do another VTherm.

That makes awful code with repeated lines (10 times...). It was not very nice, but now it is awful.
So my opinion as now is that I will not merge this PR. Sorry for your work and my time to make this response.

What could be fine is to find a way to add dynamically some underlying. But I think it is very complex

pounard:
I understand your point, nevertheless, it's somewhat really inconvenient to have more than one VTherm for the same room and I'd really like to have only one. To be honest, I'm quite happy to be able to contribute it and not let the burden all onto yourself, that's one of the great benefit I see in open source software!

How about a compromise? Let's limit the number of added heaters, say 2 for example, and in exchange, I'll try to focus two things:
 - Attempt to factorize the list whenever possible.
 - Find meaningful unit tests to add for it.

What do you think about this?

jmcollin78:
> How about a compromise? Let's limit the number of added heaters, say 2 more for example, to avoid huge list propagation, and in exchange, I'll try to focus two things:

Yes exactly was I was thinking after a full night of rest. Let's limit to 5. It seems a good compromise and a "round" number.
5 for all VTherm type. Can you go in this direction please ? Then I will do a full review and do some response to your points.
I think you miss translations (don't remember exactly if there is an impact).

And once againg, so many thanks for your contribution. That is so rare, I have to take care of contributors 👍

pounard:
> Yes exactly was I was thinking after a full night of rest. Let's limit to 5. It seems a good compromise and a "round" number.
> 5 for all VTherm type. Can you go in this direction please ?

I'm OK with this, mostly because it's enough for me :sweat_smile: but I hope there will not be people asking for more too soon.

I'll fix my PR and come back to you, thanks for showing interest in this.

jmcollin78:
> Ho! And I'll buy you a beer if you find your way to Nantes, FR sometime 😄

Be careful, je vais venir ! (Je n'y vais pas souvent mais je me le note)

pounard:
I did lower the heater number to 6, is that OK for you? I can remove one more if you prefer, yet patch is less impressive already.

tests/test_multiple_switch.py
jmcollin78: If you have 5 heaters, and the VTherm consumption is set to 100 (CONF_DEVICE_POWER: 100,) then it means that each heater can consume up to 100/5 -> 20 W.

With max_power set to 55 and current power to 50, there will be over_powering.

jmcollin78 (commented):
Hello, you forget the translations I guess. You will have to duplicate some key/value paris in the strings.json file and in all the translations/*.json file.

Keep the English words if you don't have the correct translation.

hilburn:
Is it not possible to do this as an arbitrary list? this was one of the things Better Thermostat did quite well as you'd just have 
```
underlying entities:
  - climate.one
  - climate.two
  - climate.three
  etc
 ```
rather than 
```
underlying1: climate.one
underlying2: climate.two
```

**Comment by jmcollin78** [comment]
> Is it not possible to do this as an arbitrary list? this was one of the things Better Thermostat did quite well as you'd just have
> 
> ```
> underlying entities:
>   - climate.one
>   - climate.two
>   - climate.three
>   etc
> ```
> 
> rather than
> 
> ```
> underlying1: climate.one
> underlying2: climate.two
> ```

Don't understand where your proposal is ? Are you in translations ?

**Comment by hilburn** [comment]
In better thermostat they use
```python
vol.Required(CONF_HEATER): selector.EntitySelector(
    selector.EntitySelectorConfig(domain="climate", multiple=True)
)
```
And store the entities as a list
In VTherm you have them stored as CONF_CLIMATE(_[2,3,4]), but whenever you interact with the underlying entities you wrap them in a list and iterate over them anyway - so I think it makes more sense to store them as a list in the config in the first place and remove the arbitrary limit for the number of entities that can be controlled under one VTherm
It also cleans up the config flow, as you only show currently selected entities + "add another one" rather than having 2/3 empty entries in most configurations

jmcollin78:
> It would require some code to allow migration of the current structure of CONF_<TYPE>_X -> CONF_<TYPE>_LIST but imo it's worth it

Oh yes. That is a very good way to handle this. The actual design is bad.
Do you think you will be able to give it a try ?

hilburn:
I should have some free time this week so will give it a go

jmcollin78:
> @jmcollin78 Running into some trouble with the vscode/devcontainers stuff - i have the devcontainers extension, docker installed (basically just followed the instructions in the .devcontainer README.md but I only get the attached as an error, what've I missed?
> 
> https://pastecode.io/s/h7i9woxq

If you look carefully the logs you gave me, I think the error is there:
```
[2024-10-28T13:28:13.468Z] docker: Error response from daemon: invalid mount config for type "bind": bind s
ource path does not exist: /.ssh.
See 'docker run --help'.
```
You can comment this line if you don't have one.

hilburn:
> If you look carefully the logs you gave me, I think the error is there:
> 
> ```
> [2024-10-28T13:28:13.468Z] docker: Error response from daemon: invalid mount config for type "bind": bind s
> ource path does not exist: /.ssh.
> See 'docker run --help'.
> ```
> 
> You can comment this line about .ssh if you don't have one. This is to share the login and credentials with github.

There were a couple of things wrong - i apparently did WSL wrong so had to rebuild the containers, and then when i commented out the mounts it got into the devcontainer it's not working - and none of the home assistant stuff is importing properly
image
image

jmcollin78:
@hilburn,

You should be able to run the container start command manually to see what happens.

hilburn:
Yes, got it working eventually

jmcollin78:
Hello @pounard ,

Sorry for your PR, but @hilburn does something great by allowing to add arbitrary number of underlyings. This is in the PR https://github.com/jmcollin78/versatile_thermostat/pull/590.

Thank you anyway and don't hesitate to propose great changes !

pounard:
@jmcollin78 and how about merging mine until the other one is complètes?

jmcollin78:
Hello,

I will be fast (this week) to merge if all is alright.
Just create 2 VTherms if you can't wait.

jmcollin78:
Relased with: https://github.com/jmcollin78/versatile_thermostat/releases/tag/6.6.0

Issue #576, opened 2024-10-23
Ignore set temperature changes from underlying if outside of configured min/max range


gpayer:
Any set temperature changes sent by an underlying should be filtered by using the configured min/max temperature values.
This way VTherm usage via the HA UI and via an underlying behave the same way.

**Is your feature request related to a problem? Please describe.**
And additionally stupid behaviour (like sending a set temperature of 5.0 °C if switched off) by underlyings (looking at you, Bosch BTH-RA) is ignored.

**Describe alternatives you've considered**
Abusing the light auto-correction, so set temperature changes from underlyings are ignored. However this might lead to suboptimal heating behaviour.

Originally I had planned to implement this myself, as this is a pretty minor change, but I simply can't find the time to do so.

jmcollin78:
Hello @gpayer ,

Yes this is a nice suggestion. I would love a PR for doing that because, my time is also so short.
I could help if needed, i seems to be easy to implement.

jmcollin78:
Thank you @gpayer !
You are now a member of the family contributor: https://github.com/jmcollin78/versatile_thermostat/releases/tag/6.4.1

PR #581, opened 2024-10-25
Filter out-of-range target temperature sent from underlying climate devices

gpayer:
My initial goal was to filter out stupid data from my Bosch BTH-RA thermostat. But this change makes also sense in the grand scheme of things, because it unifies the user experience. It doesn't matter whether you change the target temperature via the HA frontend or directly with an underlying, min and max are considered in both scenarios.

At the moment there are no unit tests, as I still have to figure out how to create one for this scenario, but I will try to do that in the next days.

This implements #576.

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -719,6 +719,22 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
             )
             return
 
+        # Forget event when the new target temperature is out of range
+        if (
+            not new_target_temp is None
+            and not self._attr_min_temp is None
+            and not self._attr_max_temp is None
+            and (new_target_temp < self._attr_min_temp or new_target_temp > self._attr_max_temp)
+        ):
+            _LOGGER.debug(
+                "%s - underlying sent a target temperature (%s) which is out of configured min/max range (%s / %s). The value will be ignored",
+                self,
+                new_target_temp,
+                self._attr_min_temp,
+                self._attr_max_temp,
+            )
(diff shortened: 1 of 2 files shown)


custom_components/versatile_thermostat/thermostat_climate.py
```diff
@@ -719,6 +719,22 @@ async def end_climate_changed(changes: bool):
             )
             return
 
+        # Forget event when the new target temperature is out of range
+        if (
+            not new_target_temp is None
+            and not self._attr_min_temp is None
+            and not self._attr_max_temp is None
+            and (new_target_temp < self._attr_min_temp or new_target_temp > self._attr_max_temp)
```
jmcollin78: This can be safely simplified with:

and self._attr_max_temp < new_target_temp < self._attr_min_temp

This is a specific Python syntax !

jmcollin78 (commented):
Hello @gpayer ,

nice to see you take time to do the PR. I think there is a unit test failing now. Can you have a look please ?

custom_components/versatile_thermostat/thermostat_climate.py
gpayer: Thanks for the hint, now this line is pretty!

gpayer:
Hey @jmcollin78,

thanks for taking a look at my pull request. I fixed the broken test case, it was directly affected by my change, as it tried to set the temperature to a now filtered value. Now a different temperature is used and all is well.
Based on that test case I was able to create my own one, so now this code is actually tested!

jmcollin78 (approved):
That is a great change. Thank you for this. It will be in the next release.

Issue #522, opened 2024-10-05
[Feature request] - Ability to disable presets. 


tomtom13:
Request is pretty simple - ability to disable all the preset stuff. Some people have different approach to managing heating and those presets are just confussing non-technical users.

It would be nice to also disable the modes.

All of those in a sense of NOT exposing this to HA so those are not automatically harvested by HASS.

tomtom13:
Yeah, maybe don't close the issue if you've not confirmed with the reporter that this actually work.
You can't set value of 0 if you set minimum allowed temp higher than 0. Sorry, but one setting of temp ranges covering UI and backend configuration is "suboptimal".

jmcollin78:
Sorry to close the issue to rapidly, before 6.3.0 it was possible to specify 0 at temperature which deactivate the preset.

It is now not possible, that is right. So I have no way now to disable a preset.
I will think about this.

tomtom13:
I would suggest a temp setting at the central place - ie "central config" and a secondary min / max setting in thermostat place. That way you can allow global configuration, and then allow people to set per thermostat settings. 
Or simply:
- allow whatever temp user wants for presets
- allow disabling presets via configuration (not through the device or other funky kung-fu)

jmcollin78:
no activity and no real need. I close.

tomtom13:
Yeah, you need to allow minimum temp globally to be 0 - otherwise, you can't set 0 degree for preset to dissable it. I've pointed it out, but was more or less dismissed.

jmcollin78:
> I have been using versatile thermostat for over a year -- thanks for great functionality!
> 
> Now a year later I added another thermostat and tried to set temperature to 0 on two of the presets mode to disable them -- but it looks like they are still present. Is there some other way? For the other thermostats I have two two preset -- comfort and frost.

No it is not possible. A developper starts but stops. This is not so easy with VTherm architecture. Some wants more presets, some wants less presets, ... One user, one need.

The best to do is to make the preset list configurable but this is much complicated for now.

PR #610, opened 2024-11-06
Initial attempt to add disabling of presets to config_flow

hilburn:
I thought I'd open the PR just so it can be discussed
Related to #522 

TODO:
1. Load selected presets into BaseThermostat._attr_preset_modes
2. Calculate the presets/presets_away based on 1. rather than use static
3. Ensure all the various number entities load properly
4. Add translations
5. Add tests

### DIFF
--- a/custom_components/versatile_thermostat/config_flow.py
+++ b/custom_components/versatile_thermostat/config_flow.py
@@ -235,6 +235,14 @@ class VersatileThermostatBaseConfigFlow(FlowHandler):
                 except ServiceConfigurationError as err:
                     raise ServiceConfigurationError(conf) from err
 
+        motion_presets = [CONF_MOTION_PRESET, CONF_NO_MOTION_PRESET]
+        for conf in motion_presets:
+            preset = data.get(CONF_MOTION_PRESET, None)
+            if preset and (
+                preset not in data.get(CONF_ACTIVE_PRESETS, CONF_PRESETS_SELECTIONABLE)
+            ):
+                raise InvalidPresets(conf)
+
     def check_config_complete(self, infos) -> bool:
         """True if the config is now complete (ie all mandatory attributes are set)"""
         is_central_config = (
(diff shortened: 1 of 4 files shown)


jmcollin78:
Thank you for this, I will have a check later (I will be off for few days)

hilburn:
It's nowhere near done but hopefully I'll get a chance before next week to do some more

hilburn:
Yeah basically - Frost Protection, Boost, Comfort and Eco can all be individually disabled in both central config and independent VTherms
There's a new config error if you select a disabled preset for motion (I couldn't find a nice way to remove the preset from the selectors) 
I should be able to tack on to that logic of termperature > 0 to disable the presets in each, I've already changed the logic for generating the TemperatureNumber entities to not generate them for disabled presets

hilburn:
I'd have to look at it, but probably!
I think the way the presence sensor works currently is that internally there needs to be an "away" variant, but it can be hidden from the user and just take the same value as "Frost"

jmcollin78:
> I'd have to look at it, but probably! I think the way the presence sensor works currently is that internally there needs to be an "away" variant, but it can be hidden from the user and just take the same value as "Frost"

Exactly, the genericity needs to have a Frost_away. TO remove it we must add an exception and an ugly "if frost". But maybe it worth it.

hilburn:
> > I'd have to look at it, but probably! I think the way the presence sensor works currently is that internally there needs to be an "away" variant, but it can be hidden from the user and just take the same value as "Frost"
> 
> Exactly, the genericity needs to have a Frost_away. TO remove it we must add an exception and an ugly "if frost". But maybe it worth it.

It should be possible to add the exception here: https://github.com/jmcollin78/versatile_thermostat/blob/main/custom_components/versatile_thermostat/base_thermostat.py#L1385 - just make that:

```python
if not self._presence_on or self._presence_state in [
                None,
                STATE_ON,
                STATE_HOME,
            ] or preset_mode == 'frost':
```

jmcollin78:
Hello @hilburn ,

Do you think you wish to continue this PR ?

hilburn:
I do yes
Sorry I had a bad run of nonsense, ill, then my PC died (the pump in the watercooling loop gave up), and now Christmas stuff with family.
I'm going to give this another go when I get back home in a few days

jmcollin78:
Take your time, there is really no urgency. Many thinks have changed. You will have a big rebase to do I think.

hilburn:
Yup. Been looking over the 7.0 changes. Should be manageable though

jmcollin78:
Hello @hilburn , sorry but I can't keep an open PR for too long. May I close it ?
I guess there no real need on that and I don't want to complexify if there is no need.

PR #632, opened 2024-11-14
Implementing max_on_percent setting

ms5:
Implemented the `max_on_percent` expert level config to clamp the max ON time to a certain percentage of the cycle length. As discussed here https://github.com/jmcollin78/versatile_thermostat/discussions/602. 

`max_on_percent` can be set in the `configuration.yaml` like this: 

```
versatile_thermostat:
    max_on_percent: 0.8
```

I have also added tests, to check if the calculation is done correctly. 

To avoid confusions, I also exposed the following attributes to HA:
```
max_on_percent
calculated_on_percent
```
Together with the already existing `on_percent` (which no can be clamped), all the informations are available to properly debug something if needed.

### DIFF
--- a/custom_components/versatile_thermostat/prop_algorithm.py
+++ b/custom_components/versatile_thermostat/prop_algorithm.py
@@ -31,6 +31,7 @@ class PropAlgorithm:
         cycle_min: int,
         minimal_activation_delay: int,
         vtherm_entity_id: str = None,
+        max_on_percent: float = None,
     ) -> None:
         """Initialisation of the Proportional Algorithm"""
         _LOGGER.debug(
@@ -78,6 +79,7 @@ class PropAlgorithm:
         self._off_time_sec = self._cycle_min * 60
         self._security = False
         self._default_on_percent = 0
+        self._max_on_percent = max_on_percent
 
     def calculate(
         self,
@@ -160,6 +162,15 @@ class PropAlgorithm:
                 self._calculated_on_percent,
             )
             self._on_percent = self._calculated_on_percent
+
+        if self._max_on_percent is not None and self._on_percent > self._max_on_percent:
+            _LOGGER.debug(
+                "%s - Heating period clamped to %s%% (instead of %s%%) due to max_on_percent setting.",
+                self._vtherm_entity_id,
+                self._max_on_percent,
+                self._on_percent,
+            )
+            self._on_percent = self._max_on_percent
 
         self._on_time_sec = self._on_percent * self._cycle_min * 60
 
--- a/custom_components/versatile_thermostat/prop_algorithm.py
+++ b/custom_components/versatile_thermostat/prop_algorithm.py
@@ -165,7 +165,7 @@ class PropAlgorithm:
 
         if self._max_on_percent is not None and self._on_percent > self._max_on_percent:
             _LOGGER.debug(
-                "%s - Heating period clamped to %s%% (instead of %s%%) due to max_on_percent setting.",
+                "%s - Heating period clamped to %s (instead of %s) due to max_on_percent setting.",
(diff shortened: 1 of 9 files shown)


jmcollin78:
Hello @ms5,

Thank you for this, I will have a look until this week-end. I'm not at home this end of week.

custom_components/versatile_thermostat/prop_algorithm.py
```diff
@@ -161,6 +163,15 @@ def _calculate_internal(self):
             )
             self._on_percent = self._calculated_on_percent
 
+        if self._max_on_percent is not None and self._on_percent > self._max_on_percent:
+            _LOGGER.debug(
+                "%s - Heating period clamped to %s%% (instead of %s%%) due to max_on_percent setting.",
```
jmcollin78: on_percent is not a percentage (yes I know ...). Juste remove the % sign.

jmcollin78 (commented):
This is fine. Just a typo.
You will have to update the strings.json and en.json (I will do the Fr), and please update the expert_mode in README (I will report after in README-fr).

Then I will merge and publish a beta so that, you will be able to test IRL.

custom_components/versatile_thermostat/prop_algorithm.py
ms5: Thanks for the review! 

> This is fine. Just a typo.

I have already fixed the "%" typos. 

> , and please update the expert_mode in README (I will report after in README-fr).

Currently the expert mode section is related to "thermostat over climate" section, which does not fit the new setting. My suggestion is to put the info near the end of the configuration section after "Control of a central boiler" and before "Parameter summary". What do you think about this? 

> You will have to update the strings.json and en.json (I will do the Fr)

I am blanking a bit on this. My understanding is that `strings.json` holds frontend "text" that need to be translated. My change does not have any frontend exposure, expect of the attributes which can be visible in the debug menu of HA, but I don't think these should be translated. Can you please point me in the right direction on this?

jmcollin78: > Currently the expert mode section is related to "thermostat over climate" section, which does not fit the new setting. My suggestion is to put the info near the end of the configuration section after "Control of a central boiler" and before "Parameter summary". What do you think about this?

You are right. That is fine. And if you can move the over_climate expert § into this chapter. If you want, I can do it.

>  My understanding is that strings.json holds frontend "text" that need to be translated. My change does not have any frontend exposure

Once more you are totally right. Forget this comment.

ms5:
I have updated the README by creating a new Expert Mode section (as discussed), moved the existing expert mode settings there and added the new max_on_percent setting. Please have a look and tell me if you would like to have some changes. 

Have a nice Sunday 🌞

jmcollin78 (approved):
Nice work. Thank you !

jmcollin78:
Merged ! Will be in the next release.

ms5:
Awesome! Thank You!

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/releases/tag/6.8.0.beta3

You can isntall this preview release if you want. Give me some feedback if all is alright for you.

ms5:
I just had a look, but it seems the changes are not merged into the 6.8.0.beta3 tag.

jmcollin78:
You are right. I will fix that. I forgot the rebase the branch.

jmcollin78:
Should be fixed : https://github.com/jmcollin78/versatile_thermostat/releases/tag/6.8.0.beta4

ms5:
perfect! now it works. seems good so far.

ms5:
I just want to report back since I have tested this now over the last week and I did not encounter any problems.

PR #635, opened 2024-11-14
Fix side effect of out-of-range temperature filtering

gpayer:
At the moment if a change event from an underlying has an out-of-range temperature, then the whole event is ignored. However this can lead to inconsisent states as relevant changes from the underlying might be ignored, i.e. `hvac_action`.

In this case the underlying will be in state `off/idle`, while the vtherm will stay in state `off/heating`. This will resolve itself eventually when another type of event comes in, e.g. from a temperature sensor. But until then it is confusing and annoying.

With this change just the detected target temperature change is filtered out, the other fields of the new underlying state can still be processed.

I'll add or change an unit test to cover this case, but won't have time for it until tonight. So please don't merge yet :smiley:

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -725,7 +725,7 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
             )
             return
 
-        # Forget event when the new target temperature is out of range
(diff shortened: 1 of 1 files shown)


jmcollin78 (approved):
ok for me. Have you run the tests ?

jmcollin78:
Merged thank you !

jmcollin78:
I could reopen it if you prefer.

PR #658, opened 2024-11-22
Add attribute with active underlyings for easier tracking and setup

s256:
Hi everyone,

I discovered this project with the plan to improve my heating setup. I used [schedy](https://hass-apps.readthedocs.io/en/stable/apps/schedy/) before and am now looking for an alternative. 

Setting up `versatile_thermostat` was pretty easy. Not super straight forward with all the options, but obviously a lot more sophisticated compared to what I had before.

So I have a central unit (boiler) which needs to be activated by the TRVs(Climate entities) when necessary.

One thing I struggled with, was to determine which Climate Entity was keeping the Boiler going, as the plan is to only have it running, when necessary. So I thought about listing the underlyings which activate the boiler in an attribute in the existing sensor. From a user-perspective, I preferred this instead of a dedicated sensor.

Let me know what you think about this.


jmcollin78:
Hello @s256

This is great idea and very helpful. Thank you for this.

custom_components/versatile_thermostat/sensor.py
```diff
@@ -643,6 +643,14 @@ def __init__(self, hass: HomeAssistant, unique_id, name, entry_infos) -> None:
         self._attr_unique_id = "nb_device_active_boiler"
         self._attr_value = self._attr_native_value = None  # default value
         self._entities = []
+        self._attr_active_device_names = []  # Holds the names of active devices
+
+    @property
+    def extra_state_attributes(self) -> dict:
```
jmcollin78, line 659: Add the `active_device_names` into the list of `_entity_component_unrecorded_attributes` so that it will not be persisted and save place on the hard drive.

See base_thermostat.py at line 32 to have an exemple.

jmcollin78 (commented):
Almost ok

custom_components/versatile_thermostat/sensor.py
s256: Thank you for your feedback! Happy to adapt it :) 
But I am not sure if I can follow. [L32](https://github.com/jmcollin78/versatile_thermostat/blob/main/custom_components/versatile_thermostat/base_thermostat.py#L32) doesnt look right as an example.

Generally yes: we don't want to persist this on the hard drive.
My code extends the general configuration. 
So I was now looking to the list of `_entity_component_unrecorded_attributes` which is assigned to the central configuration. 
But all `_entity_component_unrecorded_attributes`  I could find are assigned to either the base thermostat (which hands down basic attributes and entities to all other climate entities) or the actual thermostats. 

(I am not a dev, so chances are pretty high, I don't understand the code ;) ) sorry :/

jmcollin78: ok. Just add this after the class declaration and it will be ok:

    _entity_component_unrecorded_attributes = SensorEntiry._entity_component_unrecorded_attributes.union(  # pylint: disable=protected-access
        frozenset(
            {
                "active_device_names"
            }
        )
    )

jmcollin78:
Hello @s256 ,

I have terminated some big changes. Can you please rebase you modifications so that I will be able to integrate it for the next release ? 
Thank you in advance

s256:
it's rebased. But I still haven't found place to add this attribute to `_entity_component_unrecorded_attributes` for the sensor entity. Or do you want me to create that for the `NbActiveDeviceForBoilerSensor` class?

custom_components/versatile_thermostat/sensor.py
jmcollin78: Add those line after the class declaration as this:
```
class NbActiveDeviceForBoilerSensor(SensorEntity):
    """Representation of the threshold of the number of VTherm
    which should be active to activate the boiler"""

    _entity_component_unrecorded_attributes = SensorEntiry._entity_component_unrecorded_attributes.union( # pylint: disable=protected-access
           frozenset( { "active_device_names" } ) )

    def __init__(self, hass: HomeAssistant, unique_id, name, entry_infos) -> None:

jmcollin78:
Hello, i will merge and finish the dev myself. I need your feature for a bug. Thank you for your PR.

jmcollin78 (approved):

PR #659, opened 2024-11-23
Added warning about auto start/stop and air-to-air heat pumps

carlolars:
Yes this happened to me :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,6 +433,9 @@ Once configured, you will now have a new entity of type `switch` that allows you
 
 The detection algorithm is described [here](https://github.com/jmcollin78/versatile_thermostat/issues/585).
 
+> [!WARNING]
+> Using the Auto start/stop feature on an Air-to-Air heat pump when the outside temperature drops below freezing can prevent the heat pump from running its defrost-cycle correctly, resulting in an outdoor unit that is covered in ice.
+
 ### For a thermostat of type ```thermostat_over_valve```:
 ![image](images/config-linked-entity3.png)
 You can choose up to domain entity ```number``` or ```ìnput_number``` which will control the valves.


jmcollin78:
Hello,

The auto-start/stop only turn-off the VTherm but the unit is autonomous to do the unfreeze cycle not ?

carlolars:
> Hello,
> 
> The auto-start/stop only turn-off the VTherm but the unit is autonomous to do the unfreeze cycle not ?

For my unit, when it is turned off it stops everything so unfortunately it won't do a defrost cycle.

README.md
```diff
@@ -433,6 +433,9 @@ Once configured, you will now have a new entity of type `switch` that allows you
 
 The detection algorithm is described [here](https://github.com/jmcollin78/versatile_thermostat/issues/585).
 
+> [!WARNING]
+> Using the Auto start/stop feature on an Air-to-Air heat pump when the outside temperature drops below freezing can prevent the heat pump from running its defrost-cycle correctly, resulting in an outdoor unit that is covered in ice.
+
```
jmcollin78, line 438: I would like to avoid too much fear for the reader. Maybe we can write it like that:

Some rare Air-to-Air pump don't ... Please check if you are in that case before using the auto-start/stop feature.

Typically mine don't do that. What if you go on holidays for two weeks in winter ? You have to keep you heat pump on during all the days ? I think your point is very strange.

carlolars: Yes I agree that the wording was a little bit strong. And it might just be old heat pumps that work in that way, mine is 10yrs+.

Lets just wait and see, if others start having problem with frozen outdoor units then you'll have a hint on what the problem is.

jmcollin78: Yes. Thank you for the tips anyway. Can you close the PR then ?

carlolars:
(Feel free to modify/close or do what you like with this PR)

PR #686, opened 2024-12-05
Update README-fr.md

eflye:
Align plotly-graph for fr and en documentation


jmcollin78:
Done in documentation rework !

PR #769, opened 2025-01-02
(chore) Lint base_* files

cova-fe:
Removes typos and linter warnings for lines too long. Line lenght is 88, as per black default.

### DIFF
--- a/custom_components/versatile_thermostat/base_entity.py
+++ b/custom_components/versatile_thermostat/base_entity.py
@@ -3,7 +3,8 @@
 import logging
 from datetime import timedelta
 from homeassistant.core import HomeAssistant, callback, Event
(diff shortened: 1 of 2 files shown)


custom_components/versatile_thermostat/base_entity.py
```diff
@@ -3,7 +3,8 @@
 import logging
 from datetime import timedelta
 from homeassistant.core import HomeAssistant, callback, Event
-from homeassistant.components.climate import ClimateEntity, DOMAIN as CLIMATE_DOMAIN
+from homeassistant.components.climate import ClimateEntity
+from homeassistant.components.climate.const import DOMAIN as CLIMATE_DOMAIN
```
jmcollin78, line 7: I wonder how it could works without...

jmcollin78 (approved):
Thank you for this.

PR #793, opened 2025-01-08
issue #790 - hvac-off-reason is not restored after restart

jmcollin78:
issue #790 - hvac-off-reason is not restored after restart


jmcollin78 (commented):
ok fine.

PR #799, opened 2025-01-09
Allow `coef_ext` values with up to three decimals precision

felixfischer:
This is a friendly suggestion for improving the `coef_ext` input field.

When setting my own heating, I realised that the permitted step size of `0.01` is too coarse for me. A value of 0.01 leads to the temperature being too low and 0.02 to it being too high. But it is not possible to enter a value in between! I had to use my browser's developer tools to hack the step-size attribute of the input field and after some trial and error ended up with 0.016 as my ideal value.

This change aims to fix this shortcoming by setting the step size to `0.001` and thus allowing more precise numbers.


jmcollin78:
Hello @felixfischer ,

I already have done an update on the main branch. This will be released this week-end.

Anyway thank you for the PR.

PR #905, opened 2025-02-10
Add minimal_deactivation_delay

tmad:
In some cases users may want to keep the equipment on if it will deactivate for a short time. This could help reduce wear on gas boilers.


custom_components/versatile_thermostat/prop_algorithm.py
```diff
@@ -187,6 +192,18 @@ def _calculate_internal(self):
 
         self._off_time_sec = self._cycle_min * 60 - self._on_time_sec
 
+        # Do not stop heating when off time less than xx sec
+        if self._off_time_sec < self._minimal_deactivation_delay:
+            if self._off_time_sec > 0:
+                _LOGGER.info(
+                    "%s - Heating period due to heating period too small (%f < %f)",
```
jmcollin78: I think this message is wrong.
We should maybe set the message like this: "%s - Force 100% heating cause off (%f) period is under minimal_deactivation_delay (%)", self._vtherm_entity_id,  self._off_time_sec, self._minimal_deactivation_delay

custom_components/versatile_thermostat/translations/en.json
```diff
@@ -199,6 +199,7 @@
         "description": "Configuration of advanced parameters. Leave the default values if you don't know what you are doing.\nThese parameters can lead to very poor temperature control or bad power regulation&nbsp;[![?](https://img.icons8.com/color/18/help.png)](https://github.com/jmcollin78/versatile_thermostat/blob/main/documentation/en/feature-advanced.md)",
         "data": {
           "minimal_activation_delay": "Minimum activation delay",
+          "minimal_deactivation_delay": "Minimal deactivation delay",
```
jmcollin78, line 202: This translation should added 2 times (I know this is bad). Once in the option section below.

jmcollin78, line 202: Can you add this text in the fr.json file with a big TODO so that I will think to replace it ?

Thank you

jmcollin78 (commented):
Hello @tmad,

You have done an amazing work to add this parameter and some test ! Thank you for this. 

I will be happy to merge it. Just some small comments and we are good.

custom_components/versatile_thermostat/prop_algorithm.py
tmad: Good catch, thanks

custom_components/versatile_thermostat/translations/en.json
tmad: Added

tmad: I've added those in fr.json, though I'm not sure where to add the TODO so it's visible and keeping a valid syntax

tmad:
Thanks for the review! I've addressed the comments.

jmcollin78:
Thank you  @tmad !
Very good work for a first PR

jmcollin78:
For your information I have fix some test issue.

PR #988, opened 2025-04-04
the German translation

FMainz:
German translation for versitale thermostat


jmcollin78:
Hello, thank you for this !

PR #1052, opened 2025-07-01
German documentatition files

FMainz:
New german documentatition files


FMainz (commented):
Updating

jmcollin78 (approved):
Thank you so much for this !

PR #1082, opened 2025-09-16
Add entity_category for all entities except climate

### LINKED ISSUE #1080: [Feature Request] - Set entity_category for configuration and diagnostic entities
**Is your feature request related to a problem? Please describe.**
- For auto-generated dashboards based on an area all non-config/diagnostic entities are added, which are all in this case. This leads to much clutter.
- If running switch actions based on areas it also changes switches which are more settings-style things.

**Describe the solution you'd like**
Set the appropriate [category](https://developers.home-assistant.io/docs/core/entity/#registry-properties) for all such entities. In my opinion only the climate entity would remain, as all others are usually only needed for diagnostics or are settings.

**Describe alternatives you've considered**
None.

**Additional context**
This would also hide the entities from the logbook, which would also be a plus in my opinion.

Thanks for this nice component!

malteger:
This assumes that in general all non climate entities are either DIAGNOSTIC or in the case of numbers and switches CONFIG entities, as the category is set in the custom base classes.

I have applied the patch to my own installation, but some more testing might be required. I do not know if the current approach by setting this in the base class is the best, but as only the climate entities should be primary this would ensure all relevant entities are affected.

Closes #1080

### DIFF
--- a/custom_components/versatile_thermostat/base_entity.py
+++ b/custom_components/versatile_thermostat/base_entity.py
@@ -5,6 +5,7 @@ from datetime import timedelta
 from homeassistant.core import HomeAssistant, callback, Event
 from homeassistant.components.climate import ClimateEntity
 from homeassistant.components.climate.const import DOMAIN as CLIMATE_DOMAIN
+from homeassistant.const import EntityCategory
 from homeassistant.helpers.entity_component import EntityComponent
 from homeassistant.helpers.entity import Entity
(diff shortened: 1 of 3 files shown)


jmcollin78 (approved):
Thanks for this implementation. All seems fine for me.

jmcollin78:
There is some error in the unit tests but don't worry, it is not your fault. I will to see what goes wrong since 2025.09.

PR #1085, opened 2025-09-20
(fix) Include min_opening_degree in new opening/closing degree calculation

enryIT:
If the user properly set a different value than `0` for the min_opening_degree for a specific underlay, it should always be used when calculating the closing degree.

### DIFF
--- a/custom_components/versatile_thermostat/underlyings.py
+++ b/custom_components/versatile_thermostat/underlyings.py
@@ -1108,11 +1108,11 @@ class UnderlyingValveRegulation(UnderlyingValve):
         self._max_offset_calibration: float = None
         self._min_opening_degree: int = min_opening_degree
 
-    def _normalize_opening_closing_degree(self, opening: float) -> float:
+    def _normalize_opening_closing_degree(self, opening: float, min_opening: float) -> float:
         """Issue #902 - Normalize the opening and closing degree
         Issue #927 - Cancel the normalization"""
 
-        new_opening = max(opening, 0) if self.has_closing_degree_entity else opening
+        new_opening = max(opening, min_opening)
         new_closing = max(self._max_opening_degree - new_opening, 0) if self.has_closing_degree_entity else 100
 
         return new_opening, new_closing
@@ -1156,7 +1156,7 @@ class UnderlyingValveRegulation(UnderlyingValve):
             self._percent_open = 0
 
         # Send closing_degree if set
-        opening_degree, closing_degree = self._normalize_opening_closing_degree(self._percent_open)
+        opening_degree, closing_degree = self._normalize_opening_closing_degree(self._percent_open, self._min_opening_degree)
         # We should not change the _percent_open because it is used to check if value has bchanged
         # self._percent_open = opening_degree
 


jmcollin78 (commented):
Hello, thank you for this.
I wonder if it is right:
1. min_opening_degree is already taken into account in self._percent_open. See the code just above,
2. what issue are your trying to fix. You explain the change but not why this change. I can't figure out what was wrong and what you are trying to achieve.

I have test around this feature, and I pretty sure the min is well taken into account when calculating the closing degree.

enryIT:
> Hello, thank you for this. I wonder if it is right:
> 
> 1. min_opening_degree is already taken into account in self._percent_open. See the code just above,
> 2. what issue are your trying to fix. You explain the change but not why this change. I can't figure out what was wrong and what you are trying to achieve.
> 
> I have test around this feature, and I pretty sure the min is well taken into account when calculating the closing degree.

I'm sorry, I was wrongly supposing the issue was obvious.

The problem appear to be very simple to me, as you pointed out, the value of min_opening_degree was already taken into account in the code, but only if the calculated percent was more than 0
```
        # Caclulate percent_open
        if self._percent_open >= 1:
            self._percent_open = round(
                self._min_opening_degree
                + (self._percent_open
                   * (100 - self._min_opening_degree) / 100)
                )
        else:
            self._percent_open = 0
```

The problem I'm trying to solve is that, if the user set a minimum opening degree, it should be always taken into account; If I set it to `30` I expect it to change the actual valve range from 0-100 to 30-100, so, even if the climate entity is idling, the valve should be opened at least 30%.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -1156,7 +1156,7 @@ async def send_percent_open(self, _: float = None):
             self._percent_open = 0
```
jmcollin78: I guess the fix is to be done here: 

self._percent_open = self._min_opening_degree (and not 0 hardcoded).

What we should do is rather to put all calculation in the _normalize_opening_closing_degree function and not partially in the calling method and in the _normalize_opening_closing_degree. As is it is not easy to find out what happens.

Can you do that please ?

If not clear, I wonder if putting line 1148-1156 into the _normalize_opening_closing_degree is better.

Please check that unit tests are ok with that change.
If too complex for you, let me know and I will take back the PR.

Anyway thank you for your PR, it is helpfull.

enryIT:
I just made a big mistake with the initial pull request, I also edited the title for better understanding.

Let me recap a little bit:

- Apply floor in `send_percent_open()` also when requested == 0, clamp to `min_opening_degree` when > 0, keep the original mapping from `[min_opening_degree` to 100.
- Add `UnderlyingValveRegulation.turn_off()` override to always call `send_percent_open()`, so the `min_opening_degree` is enforced even when HVAC is OFF, otherwise it would be skipped.
- Make `is_device_active()` depend on the heat request (`thermostat.valve_open_percent` > 0,   and `hvac_mode` != OFF) instead of the physical opening; this keeps `hvac_action` OFF/IDLE when there is no demand, even if the valve is physically held at the minimum opening.

Why:
Previously, the base `turn_off()` called `send_percent_open()` only when `is_device_active()` was True; when False it skipped sending, so the valve could remain at 0% despite the floor being enforced only at send time. Now the floor is guaranteed whenever we send, including OFF, and the logical state reflects demand rather than the physical floor.

Behavior changes:
- The valve never closes below min_opening_degree.
- With no heat request, entities report OFF/IDLE while the valve may still be physically >0.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -1250,11 +1251,20 @@ async def start_cycle(
...
+
     @property
     def is_device_active(self):
         """If the opening valve is open."""
         try:
-            return get_safe_float(self._hass, self._opening_degree_entity_id) > 0
+            #return get_safe_float(self._hass, self._opening_degree_entity_id) > 0
+            if self.hvac_mode == HVACMode.OFF:
+                return False
+            return self._thermostat.valve_open_percent > 0
```
jmcollin78, line 1267: When doing that you check if the valve should be open and not if it is really open. I'm not fan of this change.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -1250,11 +1251,20 @@ async def start_cycle(
         # if force:
         await self.set_valve_open_percent()
 
+    @overrides
+    async def turn_off(self):
+        _LOGGER.debug("%s - ValveRegulation turn_off -> enforce min_opening_degree", self)
+        self._percent_open = 0
+        await self.send_percent_open()
```
jmcollin78, line 1258: Send only if the valve is > min to avoid transmitting values when not necessary and save some battery.

Is the upper class turn_off not right ?

```
    async def turn_off(self):
        _LOGGER.debug("%s - Stopping underlying valve entity %s", self, self._entity_id)
        # Issue 341
        is_active = self.is_device_active
        self._percent_open = self.cap_sent_value(0)
        if is_active:
            await self.send_percent_open()
```

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -1153,7 +1154,7 @@ async def send_percent_open(self, _: float = None):
                    * (100 - self._min_opening_degree) / 100)
                 )
         else:
-            self._percent_open = 0
+            self._percent_open = self._min_opening_degree
```
jmcollin78, line 1157: Fine !

custom_components/versatile_thermostat/underlyings.py
jmcollin78, line 1267: May be we can consider the valve open is valve_open_percent is > min and not > 0

custom_components/versatile_thermostat/underlyings.py
enryIT: I think the base `UnderlyingValve.turn_off()` is not right for my goal (send only if current opening > min), because:

1. It sends only if `is_device_active()` is True (i.e., opening > 0, not > min)
If the valve is already at the min_opening (e.g., 10%), it’s still > 0, so it will send anyway, this results in useless traffic/battery drain

2. If HA temporarily thinks the valve is inactive (no state / unavailable), the base method skips sending completely, and you might miss enforcing the min_opening

custom_components/versatile_thermostat/underlyings.py
enryIT: That would make small but real demands (e.g., request=4, min=10) look inactive, that’s the opposite of what we want: a non-zero request should be “heating” even if the min_opening forces a higher physical opening.
The min_opening is a plumbing/safety constraint, not a demand threshold.

enryIT: You're right, maybe we should:

1. restore the old physical check 
```
@property
def is_device_active(self):
    try:
        return get_safe_float(self._hass, self._opening_degree_entity_id) > 0
    except Exception:
        return False
```

2. add an inner property hvac_action in `UnderlyingValveRegulation` class
```
@property
def hvac_action(self) -> HVACAction | None:
    if self.hvac_mode == HVACMode.OFF:
        return HVACAction.OFF
    try:
        if self._thermostat.valve_open_percent > 0:
            return HVACAction.HEATING
    except Exception:
        pass
    return HVACAction.IDLE if self.is_device_active else HVACAction.OFF
```

This way we keep `is_device_active` physical (truth about the hardware) and `hvac_action` for valve regulation reflect demand (any >0 request means “heating”)

custom_components/versatile_thermostat/underlyings.py
enryIT: I'll try to refactor it and test it in my environment.

custom_components/versatile_thermostat/underlyings.py
jmcollin78: I'm not confortable with the point 2. I'm pretty sure I will have issues with "the valve is open, but the VTherm is not heating".
But I have to think more about your proposal. This is a very sensible combinaison which gives me many issues.
I need more time to evaluate the changes. Maybe I will have time this week-end.

jmcollin78:
Hello @enryIT. I do something similar on VTherm v8.xx. Can you have a look please ?
And close or finish this PR please. Too long PR are not good for everyone.

enryIT:
> Hello @enryIT. I do something similar on VTherm v8.xx. Can you have a look please ? And close or finish this PR please. Too long PR are not good for everyone.

Hi @jmcollin78, I can confirm the basic functionality, such as the `min_opening` as plumbing/safety constraint and not a demand threshold, is there and is working.

I think there are two main problems with the v8 implementation:
1. If you define `min_opening` then the `hvac_action` is always `heating`, even when it shouldn't (eg. current temperature > target temperature). Small but real demands now looks active, that's what we want, but when the request is 0  the device should be idle regardless the `min_opening`
```
Example #1:
request=4
min_opening=10
then `hvac_action` should be `heating`

Example #2:
request=40
min_opening=10
then `hvac_action` should be `heating`

Example #3:
request=0
min_opening=10
then `hvac_action` should be `idle` even if the physical is heating
```

2. The `min_opening` shouldn't be taken into account if the window is open, if the window is open then the valve should totally close (maybe there are case where you want this, maybe we should have the option to account or not `min_opening`)

jmcollin78:
Hello,

> If you define min_opening then the hvac_action is always heating, even when it shouldn't

Not on my side.
Capture d’écran 2025-11-19 à 17 45 59

What is you VTherm's type ? Is it a Sonoff TRVZB or equivalent (a ovr_climate with direct valve regulation) ?

enryIT:
> Hello,
> 
> > If you define min_opening then the hvac_action is always heating, even when it shouldn't
> 
> Not on my side. Capture d’écran 2025-11-19 à 17 45 59
> 
> What is you VTherm's type ? Is it a Sonoff TRVZB or equivalent (a ovr_climate with direct valve regulation) ?

image

Please check your attributes, that vtherm is heating....
```
hvac_modes:
  - heat
  - sleep
  - "off"
min_temp: 7
max_temp: 35
target_temp_step: 0.1
preset_modes:
  - none
  - frost
  - eco
  - comfort
  - boost
current_temperature: 19.2
temperature: 19
hvac_action: heating
preset_mode: comfort
hvac_mode: heat
specific_states:
  is_on: true
  last_central_mode: null
  last_update_datetime: "2025-11-19T17:54:21.979453+01:00"
  ext_current_temperature: 18.2
  last_temperature_datetime: "2025-11-19T17:48:33.284092+01:00"
  last_ext_temperature_datetime: "2025-11-19T11:44:33.960459+01:00"
  is_device_active: false
  device_actives: []
  nb_device_actives: 0
  ema_temp: 19.21
  temperature_slope: 0.12
  hvac_off_reason: null
  total_energy: 31.8
  last_change_time_from_vtherm: "2025-11-19T17:44:33.808119+01:00"
  messages: []
  is_sleeping: false
  is_recalculate_scheduled: false
configuration:
  ac_mode: false
  type: over_climate_valve
  is_controlled_by_central_mode: false
  target_temperature_step: 0.1
  minimal_activation_delay_sec: 10
  minimal_deactivation_delay_sec: 0
  timezone: Europe/Rome
  temperature_unit: °C
  is_used_by_central_boiler: false
  max_on_percent: null
  have_valve_regulation: true
  cycle_min: 5
preset_temperatures:
  frost_temp: 7
  eco_temp: 17
  boost_temp: 20
  comfort_temp: 19
  frost_away_temp: 7
  eco_away_temp: 15
  boost_away_temp: 18
  comfort_away_temp: 17
current_state:
  hvac_mode: heat
  target_temperature: 19
  preset: comfort
requested_state:
  hvac_mode: heat
  target_temperature: 19
  preset: comfort
is_presence_configured: true
presence_manager:
  presence_sensor_entity_id: input_boolean.presence
  presence_state: "on"
is_power_configured: false
is_motion_configured: false
is_window_configured: true
is_window_auto_configured: false
window_manager:
  window_state: "off"
  window_auto_state: unavailable
  window_action: window_turn_off
  is_window_bypass: false
  window_sensor_entity_id: binary_sensor.finestra_bagno_contact
  window_delay_sec: 30
  window_off_delay_sec: 30
  window_auto_open_threshold: 3
  window_auto_close_threshold: 0
  window_auto_max_duration: 30
is_safety_configured: true
safety_manager:
  safety_state: "off"
  safety_delay_min: 60
  safety_min_on_percent: 0.5
  safety_default_on_percent: 0.1
is_auto_start_stop_configured: false
is_over_climate: true
regulation_accumulated_error: 0
regulated_target_temperature: null
vtherm_over_climate:
  start_hvac_action_date: null
  underlying_entities:
    - climate.trvzb_bagno
  is_regulated: true
  auto_fan_mode: auto_fan_high
  current_auto_fan_mode: auto_fan_none
  auto_activated_fan_mode: null
  auto_deactivated_fan_mode: null
  follow_underlying_temp_change: false
  auto_regulation_use_device_temp: false
  regulation:
    regulated_target_temperature: null
    auto_regulation_mode: auto_regulation_valve
    regulation_accumulated_error: 0
vtherm_over_climate_valve:
  have_valve_regulation: true
  valve_regulation:
    underlyings_valve_regulation:
      - - number.trvzb_bagno_valve_opening_degree
        - number.trvzb_bagno_valve_closing_degree
        - number.trvzb_bagno_local_temperature_calibration
    on_percent: 0
    power_percent: 0
    function: tpi
    tpi_coef_int: 0.6
    tpi_coef_ext: 0.01
    min_opening_degrees: "15"
    valve_open_percent: 0
    auto_regulation_dpercent: 5
    auto_regulation_period_min: 5
    last_calculation_timestamp: "2025-11-19T11:45:58.633877+01:00"
friendly_name: Heat Climate Bagno
supported_features: 401
```

underlying (trvzb)
```
hvac_modes:
  - "off"
  - auto
  - heat
min_temp: 4
max_temp: 35
target_temp_step: 0.5
current_temperature: 19.1
temperature: 19
hvac_action: idle
friendly_name: TRVZB Bagno
supported_features: 385
```

jmcollin78:
If I refer to the heating icon it seems not heating but in the attribute it is heating indeed. I will fix that. Sorry it is a PR, did your PR fix this ? Even in 8.0 release ?

jmcollin78:
Ok i've got my answer, it is an too old PR now. I will close it and fix it in another one. Thank you for report.

PR #1086, opened 2025-09-22
Disable last temperature date sensors by default

malteger:
This sensors are prone to update very frequently, as per https://developers.home-assistant.io/docs/core/entity/#registry-properties set them to disabled by default.


jmcollin78 (approved):
Thank you for this !

Issue #1001, opened 2025-04-16
[Feature request] - Add hvac mode for dormant heating system


jmcollin78:
### Discussed in https://github.com/jmcollin78/versatile_thermostat/discussions/999

<div type='discussions-op-text'>

<sup>Originally posted by **AliveDevil** April 16, 2025</sup>
I'd like to see an operation mode in VTherm, when in spring/summer I turn the heater of my central gas furnace off, which opens up all valves to reduce strain on the TRV battery, valve driver and radiator valves.
Even though the central gas furnace heater is turned off, the pump will be spun up every 24 hours to keep the water in the system flowing, and reduce clogging.

The idea here is that all VTherm over Valve and VTherm over Climate (Valve Control) get an additional hvac mode "Dormant", which will set the underlying Climate hvac mode "Off", but open the underlying opening and closing percentages to the safety percentage values (or forcing them completely open). VTherms shouldn't report any other value than "0 Power Percent" while in hvac "Dormant"-mode.

Whether the Dormant-hvac mode should apply to over Switch and over Climate-entities is up for debate.
For over switch, this would perform the Turn On-action.
For over climate, this would set the underlying climate set-temperature to the configurable max temperature in heating-mode, or min temperature in AC mode.

Or, if this is out of scope, I'd like to be able to set the underlying valve percentages manually, while the TRV is set to "Off" - I have seen that despite the VTherm being Off, it constantly fights the underlying valve percentage, when set externally. Setting the VTherm to Off should send the underlying valve to 0 once, and never touch it again.</div>

jmcollin78:
I'm looking forward to implementing this rapidly.
In the meantime you have an automation here: https://github.com/jmcollin78/versatile_thermostat/discussions/1000#discussioncomment-13109045

jmcollin78:
Implementation in progress.

jmcollin78:
Heyyy,

A pre-view release is available here: https://github.com/jmcollin78/versatile_thermostat/releases/tag/7.4.0.beta2

Don't forget to update also the Versatile Thermostat UI Card to use the new sleep mode.

Comments are welcome before the stable release (next week).

AliveDevil:
I've just installed it, and it looks like the integration at least from the default "detail"-view isn't working:
```
[...]

PR #1104, opened 2025-10-02
#1001 -  add sleep mode for VTherm

jmcollin78:
#1001 - Introduce a new sleep mode for VTherms of type `over_climate`, allowing the thermostat to turn off while keeping the valve fully open. 

Don't forget to update the VTherm UI Card to use this new feature more easily/

### DIFF
--- a/.github/chatmodes/traduction.chatmode.md
+++ b/.github/chatmodes/traduction.chatmode.md
@@ -1,0 +1,5 @@
+---
+description: "Traduction du français vers l'anglais"
+tools: []
+---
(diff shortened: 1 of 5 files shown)


lwndp:
Hey! Is there any documentation about this feature? I have trouble to understand what it does and why I would want to use it. Maybe you can clarify? Thanks! :-)

jmcollin78:
Yes: https://github.com/jmcollin78/versatile_thermostat/blob/main/documentation/en/over-climate.md#specific-functions

It is very easy: it turns off the VTherm while keeping the valve 100% open. There is not much more to say about it.

Issue #1025, opened 2025-05-15
Changing centrale configuration need a restart of HA to be applied


jmcollin78:
Cf. https://forum.hacf.fr/t/versatile-thermostat-piloter-votre-chauffage-avec-la-solution-la-plus-complete/57735/290

jmcollin78:
Not sure about the fix because it was not always reproductible

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/releases/tag/7.4.0

PR #1113, opened 2025-10-05
#1025 - try to fix need restart HA after changing central configuration

jmcollin78:
#1025 - try to fix need restart HA after changing central configuration


jmcollin78 (commented):
ok

jmcollin78 (commented):
ok

jmcollin78 (commented):
ok

PR #1149, opened 2025-10-26
Update IT translation

silviopen:
File update to align it with the FR version, which appears to be more complete.

The FR translation contains the following lines (which are missing from the EN version):
8 - Description
268 - Description
412 - window_auto_open_threshold
525 - vswitch_configuration_incorrect


jmcollin78 (approved):
Thank you for this !

PR #1257, opened 2025-11-22
Issue 1220 - Change the minimum_opening degree feature

jmcollin78:
For `over_climate` with regulation by direct valve control, two new parameters are added to the existing `minimum_opening_degrees`. The parameters are now the following:
    - `opening_threshold`: the valve opening value under which the valve should be considered as closed (and then 'max_closing_degree' will apply),
    - `max_closing_degree`: the closing degree maximum value. The valve will never be closed above this value. Set it to 100 to fully close the valve when no heating is needed,
    - `minimum_opening_degrees`: the opening degree minimum value for each underlying device when ``opening_threshold` is exceeded, comma separated. Default to 0. Example: 20, 25, 30. When the heating starts, the valve will start opening with this value and will continuously increase as long as more heating is needed.

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate_valve.py
+++ b/custom_components/versatile_thermostat/thermostat_climate_valve.py
@@ -250,6 +250,7 @@ class ThermostatOverClimateValve(ThermostatOverClimate):
         async def callback_recalculate(_):
             """Callback to set the valve percent"""
             self.recalculate()
+            await self.async_control_heating(force=False)
             self.update_custom_attributes()
             self.async_write_ha_state()
 


tomtom13 (commented):
As of:
2223cb444bd4b3d3c0180997875244ef67ebd7f7
looks good,
just the test case for:
```
(1,10,100,100,0,10.9),  # 10-100 range and 1 -> starting to open
```
Thanks for addressing those promptly ;)

jmcollin78:
A big thanks to all contributors !
And a special thanks for the beers 🍻. Cheers !

(the release is coming)

PR #1271, opened 2025-11-23
fix active devices for central boiler

mread:
active devices was including those with valve closed due to open threshold

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate_valve.py
+++ b/custom_components/versatile_thermostat/thermostat_climate_valve.py
@@ -345,7 +345,12 @@ class ThermostatOverClimateValve(ThermostatOverClimate):
         if self._is_sleeping:
             return False
         else:
-            return self.valve_open_percent > 0
+            # Check actual valve opening degree, not TPI-calculated percentage
+            # The opening threshold may result in 0% even when valve_open_percent > 0
+            for under in self._underlyings_valve_regulation:
+                if under.is_device_active:
+                    return True
+            return False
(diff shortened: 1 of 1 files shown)


jmcollin78:
Of course it breaks some tests

mread:
Ah, I should probably work out how to run those

jmcollin78:
If this is hard for you, maybe i can try on my side but not for now. You will have to wait, let says one or two weeks. What did you observ that make you change this piece of code ?

mread:
No it's fine, I'll amend the PR once I've fixed them.

I observed the central boiler showing 3 active valves even though they were all closed (0%).

jmcollin78:
> No it's fine, I'll amend the PR once I've fixed them.
> 
> I observed the central boiler showing 3 active valves even though they were all closed (0%).

If they are closed, the valve_open_percent should be 0 no ? So the issue is elsewhere ? Or is it related to V 8.1 and the new 3 parameters ?

mread:
Yes, v8.1 - the valves were at 0% but active device count was still detecting them as open

mread:
image

mread:
Sorry, I can't get the mocking setup right to make those tests pass. I'm also having other problems with v8.1 wrt opening threshold not being observed and I keep finding the boiler running with all the valves closed (which this PR fixes) or nearly closed (which this PR doesn't fix). Rather than kill my boiler I'm going to rollback to 7.4.1 and come back to this when I have more time, sorry.

Issue #1285, opened 2025-11-24
The provided Dev Container does not work when running VS Code in Windows


defunes43:
When cloning the repository and trying to use the project in Windows, two errors prevent the container from running:

- The mount `source=${localEnv:HOME}/.ssh` does not exist in Windows environments

- The line ending is different in Windows and Unix systems, and the container file is not an executable when run from Windows.

jmcollin78:
You should rather just adapt the lines in your environment when you need to test as the others do.

I close because this is not a VTherm issue but we can continue the discussion in the PR comments.

PR #1286, opened 2025-11-24
Allow dev container to run in Windows

defunes43:
Fix for https://github.com/jmcollin78/versatile_thermostat/issues/1285

- The mount source=${localEnv:HOME}/.ssh does not exist in Windows environments
=> Use the concatenation of the Unix environment variable and the Windows one. Since the other one will be empty, it will yield the correct home variable for both OS (see https://code.visualstudio.com/remote/advancedcontainers/add-local-file-mount)

- The line ending is different in Windows and Unix systems, and the container file is not an executable when run from Windows.
=> Use a .gitattributes file to force the lf ending in text files, even on Windows systems (see https://code.visualstudio.com/docs/devcontainers/tips-and-tricks#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,16 +5,16 @@
         "dockerfile": "Dockerfile"
     },
     "name": "Versatile Thermostat integration",
-    "appPort": ["8123:8123"],
+    "appPort": [
+        "8123:8123"
+    ],
     // "postCreateCommand": "container install",
     "postCreateCommand": "./container dev-setup",
-
     "mounts": [
-        "source=${localEnv:HOME}/.ssh,target=/home/vscode/.ssh,type=bind,consistency=cached",
+        "source=${localEnv:HOME}${localEnv:USERPROFILE}/.ssh,target=/home/vscode/.ssh,type=bind,consistency=cached",
         // uncomment this to get the versatile-thermostat-ui-card
-        "source=${localEnv:HOME}/SugarSync/Projets/home-assistant/versatile-thermostat-ui-card/dist,target=/workspaces/versatile_thermostat/config/www/community/versatile-thermostat-ui-card,type=bind,consistency=cached"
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -5,16 +5,16 @@
...
-    "appPort": ["8123:8123"],
+    "appPort": [
+        "8123:8123"
+    ],
     // "postCreateCommand": "container install",
     "postCreateCommand": "./container dev-setup",
-
     "mounts": [
-        "source=${localEnv:HOME}/.ssh,target=/home/vscode/.ssh,type=bind,consistency=cached",
+        "source=${localEnv:HOME}${localEnv:USERPROFILE}/.ssh,target=/home/vscode/.ssh,type=bind,consistency=cached",
```
jmcollin78, line 14: Are you sure ? Why no just update the line in your env. You have to adapt that for your case.

jmcollin78 (commented):
Not sure it is a good idea it could break all developer environments.

.devcontainer/devcontainer.json
defunes43: C’est dans la doc officielle VSCode, j’ai mis un lien, donc plutôt très sûr oui.

Pour aller dans ce sens, dans https://github.com/devcontainers/spec/issues/335#issuecomment-1804901201, un dev de VSCode confirme:

"${localEnv:HOME}${localEnv:USERPROFILE}" is expected to work across Windows and Linux, as either one must be set depending on the OS.

—

J’entends l’argument sur pourquoi ne pas modifier dans mon environnement, mais c’est plus large que « mon environnement », c’est tous les utilisateurs de Windows sur leur machine de dev. 

Ça va dans le sens de la mise en place du Dev Container, n’importe qui peut cloner le repo, lancer VS Code, et se mettre au boulot

jmcollin78 (approved):
ok. Thank you !

PR #1337, opened 2025-12-01
Use unique_id to generate _attr_unique_id

spolitov:
This commit changes logic on how _attr_unique_id is generated for subdevices.
Since device_name should not be unique, using it for _attr_unique_id could result in duplicate ids.

To avoid breaking existing installations, this logic is applied only to newly added thermostats.

### DIFF
--- a/custom_components/versatile_thermostat/__init__.py
+++ b/custom_components/versatile_thermostat/__init__.py
@@ -131,7 +131,6 @@ async def async_setup(
         )
         await api.init_vtherm_links()
         await api.notify_central_mode_change()
-        await api.reload_central_boiler_entities_list()
 
     if hass.state == CoreState.running:
         await _async_startup_internal()
@@ -183,7 +182,6 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
 
     if hass.state == CoreState.running:
-        await api.reload_central_boiler_entities_list()
         await api.init_vtherm_links(entry.entry_id)
 
     return True
@@ -205,7 +203,7 @@ async def update_listener(hass: HomeAssistant, entry: ConfigEntry) -> None:
         # Reload the central boiler list of entities
         api: VersatileThermostatAPI = VersatileThermostatAPI.get_vtherm_api(hass)
         if api is not None:
-            await api.reload_central_boiler_entities_list()
+            await api.central_boiler_manager.reload_central_boiler_entities_list()
             await api.init_vtherm_links(entry.entry_id)
 
 
--- a/.devcontainer/lovelace
+++ b/.devcontainer/lovelace
@@ -8,7 +8,7 @@
       "views": [
         {
           "type": "sections",
-          "max_columns": 4,
+          "max_columns": 5,
           "title": "Synthèse",
           "path": "default_view",
           "icon": "mdi:alarm-panel",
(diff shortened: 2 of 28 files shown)


jmcollin78 (dismissed):
Hello, interesting approch. Thank you !

jmcollin78:
Can you please add a test that shows the change is working? You already have a test class for migration.
Many test are failed. Try to fix it if you can. If too difficult I can help of course 
Thank you.

spolitov:
I've fixed tests, the failures were caused by different behavior in switch.py.
The `entry` is passed to respective switch ctor.
While in sensor.py, `entry.data` is passed.

I suppose we need 2 aspects to be tested:
1) Create 2 versatile thermostats with the same name. And check that there are no conflicts.
2) Create config entry in old format and check that sensors etc. use device_name after migration.

Do you know which existing tests I could use at a starting point?

jmcollin78:
> Do you know which existing tests I could use at a starting point

Yes a good staring point is the test named full_start. It creates some basic VTherm and check the good creation. There are pretty old tests but still functionnals.

spolitov:
I've added `test_duplicate_name`.
Do you think we need to check all subentries and thermostat types?
Or single combination is enough?

jmcollin78:
I will merge after a big change for 8.3.0 which is in beta test. I don't to take a risk on the 8.3.0. But it is ok for me, just after the 8.3.0 I will merge.

Adityakk9031:
https://github.com/yugabyte/yugabyte-db/pull/28299?notification_referrer_id=NT_kwDOCI5iRbUxODI3NzkyMjU4NjoxNDM1NDg5OTc&notifications_query=repo%3Ayugabyte%2Fyugabyte-db#issuecomment-3645786953

Adityakk9031:
@spolitov please check this pr

custom_components/versatile_thermostat/__init__.py
```diff
@@ -236,7 +237,9 @@ async def async_migrate_entry(hass: HomeAssistant, config_entry: ConfigEntry):
 
         thermostat_type = config_entry.data.get(CONF_THERMOSTAT_TYPE)
 
-        if thermostat_type == CONF_THERMOSTAT_CENTRAL_CONFIG:
+        old_version = config_entry.version * 100 + config_entry.minor_version
+
+        if old_version < 201 and thermostat_type == CONF_THERMOSTAT_CENTRAL_CONFIG:
```
spolitov, line 242: BTW @jmcollin78 are you ok with this change? It is not related to the unique_id itself, but it looks strange to me that we always reset those values in case of migration.

jmcollin78: Yes it is fine !

jmcollin78:
You can install main branch if you want to test. Will do a beta later.

jmcollin78:
Hello @spolitov ,

I will have to revert your PR.
All entities are duplicated with _2 suffix. This breaks all dashboard and automation of every body.
We are almost two to have this kink of issues. @KipK  and me. Because we have some important change to merge and I don't to take any risks for the current users, i had to revert.

spolitov:
Do you have test setup which reproduces the issue?
Would be nice if I will be able to obtain its backup and debug locally.

In my tests I saw _2 suffix only in scenarios when there were vtherms with the same name, so they were generating the same entity_id prefix. But there were both entity_id w/o this suffix and with this suffix.

jmcollin78:
Hello @spolitov,

To reproduce, install the 8.2 (git checkout). Creates one VTherm. Apply your change and restart Versatile. You will have all temperature sensors grayed and new entities with _2 names.

But it seems to me obvious that if you change the `unique_id` entities will be duplicated. Do you agree with that ?
So i'm not convience it will be possible to change this without breaking all entities of all the users.

spolitov:
Hi @jmcollin78,

> To reproduce, install the 8.2 (git checkout). Creates one VTherm. Apply your change and restart Versatile. You will have all temperature sensors grayed and new entities with _2 names.

I tried with 8.3.1 which also does NOT contain this PR. And I don't see _2 in names. 
So looks like something extra should be there.


> But it seems to me obvious that if you change the `unique_id` entities will be duplicated. Do you agree with that ? 

Yes, but there is logic to don't touch `unique_id` for existing entries and apply it only to new entries.
I.e. for old entries we store device name to a separate field:
```
        if old_version < 202:
            new[CONF_UNIQUE_ATTR_ID] = config_entry.data.get(CONF_NAME)
```
and then there is logic for generating unique_id:
```
def gen_attr_uniq_id(unique_id: str, data: dict, suffix: str):
    """Returns attr_uniq_id based on specified unique_id and entry data"""
    prefix = data.get(CONF_UNIQUE_ATTR_ID, unique_id)
    return f"{prefix}_{suffix}"
```
I.e. it should use device name as a prefix for old entries and device unique id for new entries.

jmcollin78:
For me the point is not in the migration function but directly in the sensor creation where you change the unique_id. At each restart, the entity is recreated and the `unique_id` should be invariant else a new one is created.

Is that your understanding also ?

spolitov:
> For me the point is not in the migration function but directly in the sensor creation where you change the unique_id. At each restart, the entity is recreated and the `unique_id` should be invariant else a new one is created.
> 
> Is that your understanding also ?

Yes, and by design it should be `{device_name}_{entity_suffix}` for devices that were added by user before this PR.
And `{unique_id}_{entity_suffix}` for devices added by user after it.

For instance if VTherm with name some_vtherm was added with 8.3 (and before) the energy sensor will receive _attr_unique_id - "some_vtherm_energy".
But if it was added with new release, then _attr_unique_id should be "XXXXXXXX_energy", where XXXXXXXX is random string generated at a time when VTherm was added.
And it works for me this way locally.

So I suppose there is something different with your setup, that is not handled correctly and subdevice gets another _attr_unique_id.

BTW could it happen that branch that you use already has CONFIG_MINOR_VERSION increased to 2?
In this case, all devices already have version 2.2, and would be considered as a "new".

jmcollin78:
> BTW could it happen that branch that you use already has CONFIG_MINOR_VERSION increased to 2?
> In this case, all devices already have version 2.2, and would be considered as a "new".

Yes it is possible because of the auto TPI PR which touch the same migration function. I should deep dive on it if I want to reopen this change. For now, I not confident with this kind of changes (the proove).

Issue #1032, opened 2025-05-22
[Feature Request] : implement swing_horizontal_mode/swing_horizontal_modes


roumano:
## Configuration

I've create a versatile thermostat over a existing thermostat

The original thermostat have a swing_horizontal_modes :  

```
swing_horizontal_modes:
  - auto
  - 1_left
  - "2"
  - "3"
  - "4"
  - 5_right
  - split
  - swing
swing_horizontal_mode: "3"
```

The versatile climate created from it doesn't have it :  
```
swing_horizontal_modes: []
swing_horizontal_mode: ""
```

So i can't manage the swing from the GUI over the versatile climate : (i've click on the "Mode oscillation" and we can see a small empty list bellow "Mode oscillation") :  

![Image](https://github.com/user-attachments/assets/81530df7-7142-41ba-9033-6a3147cce075)

If i use the gui over the original climate, i can mange the horizontal swing :  

![Image](https://github.com/user-attachments/assets/5cd75457-0353-4883-bbef-5a30b14d35f9)

* My Full original climate attributes : 
```
hvac_modes:
  - "off"
  - heat
  - dry
  - cool
  - fan_only
  - heat_cool
min_temp: 16
max_temp: 31
target_temp_step: 0.5
fan_modes:
  - auto
  - "1"
  - "2"
  - "3"
  - "4"
  - "5"
swing_modes:
  - auto
  - 1_up
  - "2"
  - "3"
  - "4"
  - 5_down
  - swing
swing_horizontal_modes:
  - auto
  - 1_left
  - "2"
  - "3"
  - "4"
  - 5_right
  - split
  - swing
current_temperature: 23
temperature: 17.5
fan_mode: "3"
swing_mode: 5_down
swing_horizontal_mode: "3"
vane_horizontal: "3"
vane_horizontal_positions:
  - auto
  - 1_left
  - "2"
  - "3"
  - "4"
  - 5_right
  - split
  - swing
vane_vertical: 5_down
vane_vertical_positions:
  - auto
  - 1_up
  - "2"
  - "3"
  - "4"
  - 5_down
  - swing
friendly_name: Clim RDC
supported_features: 937
```

* My Full versatile climate attributes: 
```
hvac_modes:
  - "off"
  - heat
  - dry
  - cool
  - fan_only
  - heat_cool
min_temp: 16
max_temp: 31
target_temp_step: 0.5
fan_modes:
  - auto
  - "1"
  - "2"
  - "3"
[...]

PR #1372, opened 2025-12-06
feat #1032: implement swing_horizontal_mode/swing_horizontal_modes

Gamso:
Feat [#1032](https://github.com/jmcollin78/versatile_thermostat/issues/1032)
Mapping from underlying attributes.
Add tests on swing vertical/horizontal mode.

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -124,6 +124,7 @@ class BaseThermostat(ClimateEntity, RestoreEntity, Generic[T]):
         self._fan_mode = None
         self._humidity = None
         self._swing_mode = None
+        self._swing_horizontal_mode = None
         self._ac_mode = None
 
         self._cur_temp = None
@@ -339,6 +340,7 @@ class BaseThermostat(ClimateEntity, RestoreEntity, Generic[T]):
         self._humidity = None
         self._fan_mode = None
         self._swing_mode = None
+        self._swing_horizontal_mode = None
         self._cur_temp = None
         self._cur_ext_temp = None
 
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -940,6 +940,28 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
         return None
 
     @property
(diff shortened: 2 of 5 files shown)


Gamso:
Je suis pas un grand adepte du python donc si il y a des choses à revoir n'hésite pas !
J'ai pris le parti de faire un test car je voyais pas comment faire autrement pour tester le bon mapping entre le composant et le sous-jacent. (pas réussi/compris depuis le HA devcontainer).
Et quitte à tester le swing horizontal, j'ai aussi fait pour le swing vertical.

custom_components/versatile_thermostat/base_thermostat.py
```diff
@@ -1271,6 +1289,11 @@ async def async_set_swing_mode(self, swing_mode: str):
         write_event_log(_LOGGER, self, f"Set swing mode: {swing_mode}")
         return
 
+    async def async_set_swing_horizontal_mode(self, swing_horizontal_mode: str):
```
jmcollin78: Idem only in ThermostatOverClimate class.

custom_components/versatile_thermostat/base_thermostat.py
```diff
@@ -853,6 +855,22 @@ def swing_modes(self) -> list[str] | None:
         """
         return None
 
+    @property
```
jmcollin78: Hello, I think this is not necessary in the base class but only in the thermostat_climate.py. This could be safely removed I think.

custom_components/versatile_thermostat/base_thermostat.py
Gamso: Done. I was inspired by swing_mode.

custom_components/versatile_thermostat/base_thermostat.py
Gamso: Done. I was inspired by swing_mode.

jmcollin78 (approved):
Fine ! Thank you.

Gamso:
@jmcollin78 

More I read the code, and more I wonder if swing_mode shouldn't be put only in ThermostatOverClimate, as you asked me to do for swing_horizontal_mode. Or maybe I'm missing something.

jmcollin78:
Yes swing mode are only available for `over_climate` VTherm.

Issue #787, opened 2025-01-06
Auto-fan seems not work with MelCloud


Gamso:
I configured a thermostat over_climate with a Mitsubishi AC.
I used the MelCloud integration to control my AC.

The temperature control works, but I've noticed that the auto-ventilation doesn't work. It usually takes me 2 hours to gain 2-3°C in my living room, and I guess that's because the fan speed never changes. 
By default, the fan is in automatic mode (I have 5 fan speed levels: 1 to 5 + automatic mode) and it never blows hard.
Tonight, I manually changed the fan speed to 4 and the temperature rose much faster.

I have the impression that auto-fan doesn't work with MelCloud integration, but I could be wrong.

My VTherm attributes are the following:
```yaml
hvac_modes: off, heat, dry, cool, fan_only, heat_cool
min_temp: 7
max_temp: 30
target_temp_step: 0.1
fan_modes: auto, 1, 2, 3, 4, 5
preset_modes: none, frost, eco, comfort, boost
swing_modes: auto, 1_up, 2, 3, 4, 5_down, swing
current_temperature: 19
temperature: 19
fan_mode: 1
hvac_action: heating
preset_mode: comfort
swing_mode: swing
is_on: true
hvac_mode: heat
type: null
is_controlled_by_central_mode: false
last_central_mode: null
frost_temp: 7.5
eco_temp: 16
boost_temp: 22
comfort_temp: 19
frost_away_temp: 7.5
eco_away_temp: 16
boost_away_temp: 19
comfort_away_temp: 18
target_temperature_step: 0.1
ext_current_temperature: 11.7
ac_mode: true
saved_preset_mode: comfort
saved_target_temp: 19
saved_hvac_mode: heat
last_temperature_datetime: 2025-01-06T19:32:19.434308+01:00
last_ext_temperature_datetime: 2025-01-06T19:05:30.053289+01:00
minimal_activation_delay_sec: 10
total_energy: 0
last_update_datetime: 2025-01-06T19:32:37.629445+01:00
timezone: Europe/Paris
temperature_unit: °C
is_device_active: true
device_actives: climate.mitsubishi
nb_device_actives: 1
ema_temp: 18.74
is_used_by_central_boiler: false
temperature_slope: 2.92
hvac_off_reason: null
max_on_percent: null
have_valve_regulation: false
last_change_time_from_vtherm: 2025-01-06T19:04:40.418711+01:00
[...]

PR #1378, opened 2025-12-07
Improve auto fan mode to work with any fan_modes values.

Gamso:
At the beginning of the year, I opened an issue [#778](https://github.com/jmcollin78/versatile_thermostat/issues/787) regarding auto_fan_mode not being supported according to the fan_modes values.
I propose a generic implementation that should work regardless of the fan_modes values. (except for AUTO_FAN_DEACTIVATED_MODES with specific values).

I chose this logic :

**fan_modes = low, medium, high** 
|AUTO_FAN_MODE                 | fan speed|
|--------------------------------|-----------|
|CONF_AUTO_FAN_LOW         |low          |
|CONF_AUTO_FAN_MEDIUM  |medium   |
|CONF_AUTO_FAN_HIGH        |high         |
|CONF_AUTO_FAN_TURBO     |high         |

**fan_modes = low, medium, high, turbo** 
|AUTO_FAN_MODE                 | fan speed|
|--------------------------------|-----------|
|CONF_AUTO_FAN_LOW         |low|
|CONF_AUTO_FAN_MEDIUM  |medium|
|CONF_AUTO_FAN_HIGH        |high|
|CONF_AUTO_FAN_TURBO     |turbo|

**fan_modes = low, medium_low, medium, medium_high, high** 
|AUTO_FAN_MODE                 | fan speed|
|--------------------------------|-----------|
|CONF_AUTO_FAN_LOW         |medium_low|
|CONF_AUTO_FAN_MEDIUM  |medium   |
|CONF_AUTO_FAN_HIGH        |medium_high|
|CONF_AUTO_FAN_TURBO     |high|

### DIFF
--- a/custom_components/versatile_thermostat/const.py
+++ b/custom_components/versatile_thermostat/const.py
@@ -131,7 +131,7 @@ CONF_OPENING_THRESHOLD_DEGREE = "opening_threshold_degree"
 CONF_LOCK_CODE = "lock_code"
 CONF_LOCK_USERS = "lock_users"
 CONF_LOCK_AUTOMATIONS = "lock_automations"
- 
+
 CONF_VSWITCH_ON_CMD_LIST = "vswitch_on_command"
 CONF_VSWITCH_OFF_CMD_LIST = "vswitch_off_command"
 
@@ -405,7 +405,7 @@ ATTR_TOTAL_ENERGY = "total_energy"
 ATTR_MEAN_POWER_CYCLE = "mean_cycle_power"
 
 AUTO_FAN_DTEMP_THRESHOLD = 2
-AUTO_FAN_DEACTIVATED_MODES = ["mute", "auto", "low"]
+AUTO_FAN_DEACTIVATED_MODES = ["mute", "auto", "low", "quiet", "1"]
 
 CENTRAL_CONFIG_NAME = "Central configuration"
 
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -420,17 +420,58 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
             except ValueError:
                 return None
 
-        fan_modes = self.fan_modes
+        # Remove special modes like "auto"
+        fan_modes = self.fan_modes or []
+        speed_modes = [
+            mode for mode in fan_modes
+            if mode not in ["auto"]
+        ]
+
+        # We suppose speed_modes are ordered from low to high speed
+        num_speeds = len(speed_modes)
+        if num_speeds == 0:
+            self._auto_activated_fan_mode = None
+            return
+
+        # We suppose that the speed modes contains at least 3 values
+        # fan_modes = low, medium, high :
+        #    |CONF_AUTO_FAN_LOW     |low    |
+        #    |CONF_AUTO_FAN_MEDIUM  |medium |
+        #    |CONF_AUTO_FAN_HIGH    |high   |
+        #    |CONF_AUTO_FAN_TURBO   |high   |
+        # fan_modes =  low, medium, high, turbo :
+        #    |CONF_AUTO_FAN_LOW     |low  |
+        #    |CONF_AUTO_FAN_MEDIUM  |medium    |
+        #    |CONF_AUTO_FAN_HIGH    |high |
+        #    |CONF_AUTO_FAN_TURBO   |turbo   |
+        # fan_modes = low, medium_low, medium, medium_high, high :
+        #    |CONF_AUTO_FAN_LOW     |medium_low  |
+        #    |CONF_AUTO_FAN_MEDIUM  |medium      |
+        #    |CONF_AUTO_FAN_HIGH    |medium_high |
+        #    |CONF_AUTO_FAN_TURBO   |high        |
+        target_index = -1
         if auto_fan_mode == CONF_AUTO_FAN_LOW:
-            self._auto_activated_fan_mode = find_fan_mode(fan_modes, "low")
+            if num_speeds >= 4:
+                target_index = num_speeds - 4
+            else:
+                target_index = 0
         elif auto_fan_mode == CONF_AUTO_FAN_MEDIUM:
-            self._auto_activated_fan_mode = find_fan_mode(fan_modes, "mid")
+            if num_speeds >= 4:
+                target_index = num_speeds - 3
+            else:
+                target_index = 1
         elif auto_fan_mode == CONF_AUTO_FAN_HIGH:
-            self._auto_activated_fan_mode = find_fan_mode(fan_modes, "high")
+            if num_speeds >= 4:
+                target_index = num_speeds - 2
+            else:
+                target_index = 2
         elif auto_fan_mode == CONF_AUTO_FAN_TURBO:
-            self._auto_activated_fan_mode = find_fan_mode(
-                fan_modes, "turbo"
-            ) or find_fan_mode(fan_modes, "high")
+            target_index = num_speeds - 1
+
+        if target_index >= 0:
+            self._auto_activated_fan_mode = speed_modes[target_index]
+        else:
+            self._auto_activated_fan_mode = None
 
         for val in AUTO_FAN_DEACTIVATED_MODES:
             if find_fan_mode(fan_modes, val):
--- a/tests/test_auto_fan_mode.py
+++ b/tests/test_auto_fan_mode.py
@@ -24,12 +24,12 @@ from .commons import *  # pylint: disable=wildcard-import, unused-wildcard-impor
 
 @pytest.mark.parametrize("expected_lingering_tasks", [True])
 @pytest.mark.parametrize("expected_lingering_timers", [True])
-async def test_over_climate_auto_fan_mode_turbo(
+async def test_over_climate_auto_fan_mode_with_3_fan_speed_values(
     hass: HomeAssistant, skip_hass_states_is_state, skip_send_event
 ):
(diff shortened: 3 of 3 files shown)


jmcollin78:
Hello, I read your PR and I think I understand your intention. I think it is fine for me.

Ii have another user with MelCloud also that has others values like 1, 2, 3, 4. I wonder if you can do something for him because you have done this great job: https://github.com/jmcollin78/versatile_thermostat/discussions/1365

jmcollin78 (commented):
ok. i will wait your answer before approving  but it seems good for me.

Gamso:
> ok. i will wait your answer before approving but it seems good for me.

In fact I have exactly the same values like other user. I juste explain with low,medium,high because it's much easy to understand.

If you check my tests, I use 1,2,3 values on one and low,medium,high on another to guarantee that fan_mode values are not decisive.

Gamso:
I just noticed that auto_fan activation is based on a hardcoded AUTO_FAN_DTEMP_THRESHOLD value at 2°C.
I will do another PR to do it configurable.
I currently have an automation for the fan, and a 2°C difference is too large in my case.

In addition, in my automation, I do sure to never trigger the maximum level at night so as not to be woken up. I'll see if it would be relevant to do another PR on that as well. When I said “night,” I actually meant the heating starting in the morning before waking up.

jmcollin78:
> I just noticed that auto_fan activation is based on a hardcoded AUTO_FAN_DTEMP_THRESHOLD value at 2°C.
> I will do another PR to do it configurable.

I try to limit at maximum configuration option when it is not necessary. The integration is already too much complicated for users. I not sure it is important to change this and to make it configurable. More over there is no fan configuration page in which you can set this option. Not fan.

Gamso:
I understand. I'll wait until it's merged and released to see if it's sufficient for my needs. Otherwise, I'll think about a better approach to integrate my requirements without complicating the setting.

jmcollin78:
I will merge this this weekend just after the 8.3 release.

jmcollin78 (approved):
Thank you. Will integrate this in 8.3 release today

foudepc83:
Hello,

Thanks for the detailed discussion, it helped me understand the auto-fan design much better.

In my case (MelCloud / Mitsubishi), I’m not really missing a new threshold or a new advanced setting.
What I’m missing is more a *logical separation* between:

- a base fan mode (used most of the time)
- and the auto-fan boost logic (used only when ΔT is large)

Today, when auto-fan is enabled, the fan always falls back to AUTO outside of boost phases.
For PACs where the internal sensor is poorly placed, this makes the unit slow down too early.

What would fit my use case very well would be:
- keep a fixed fan mode as the default
- let auto-fan temporarily override it only during boost
- then return to the base fan mode automatically

This would avoid adding new thresholds or complex options, while still keeping the auto-fan behavior intact.

Just sharing the idea in case it fits the project philosophy.
Thanks again for the great work on 8.3 👍

foudepc83:
Auto-fan works well, but houses behave very differently depending on insulation.
Having a default fan mode allows adapting VT to the thermal behavior of the building, without changing the auto-fan logic itself.


Just to make sure my idea is clear (and to avoid any misunderstanding):

The goal is not to change the auto-fan logic itself.

Conceptually, it would be something like:

fan_default_mode  # user-chosen default (1 / 2 / 3 / 4 / auto)

if auto_fan_is_active:
    set fan = auto_activated_fan_mode   # existing logic
else:
    set fan = fan_default_mode           # instead of hardcoded "auto"

This keeps the current behavior when fan_default_mode = auto,
and only changes what happens outside of the boost phase.

I’ll wait for feedback before going further, just sharing the idea for clarity 🙂

Gamso:
Indeed, it would be better to memorize and restore the fan mode at the end of the boost.
For example, in my case, my AC must be set to fan=3 to maintain the temperature approximately (main heating for 50 m²).

The AC's automatic fan mode is clearly not ideal.
I am currently testing different algorithms to regulate the fan, but I am having trouble finding one that satisfies me.

I can try to do a PR this week-end to memorize and restore the fan mode.

foudepc83:
Yes, that’s exactly what I had in mind 👍
Memorizing the fan mode and restoring it after the boost would perfectly fit my use case as well.

Thanks a lot for looking into it, I’ll be happy to test the PR when it’s ready.

jmcollin78:
Hello @Gamso ,

> Indeed, it would be better to memorize and restore the fan mode at the end of the boost. For example, in my case, my AC must be set to fan=3 to maintain the temperature approximately (main heating for 50 m²).
> 
> The AC's automatic fan mode is clearly not ideal. I am currently testing different algorithms to regulate the fan, but I am having trouble finding one that satisfies me.
> 
> I can try to do a PR this week-end to memorize and restore the fan mode.

To do that, you should use the resquestedState (the fan mode requested by user) and currentState (the real fan mode currently applied). The calculation is done in state_manager.py. I think it is relatively easy to do and I can help if you need.

clockwe:
Hello,

Thanks for the improvement.
I think I've a problem on my Hitachi PAC, due to the assumption that fan_mode order goes from low to high.
In my case it seems that the order of fan power list is reversed ( high > medium > low > diffuse > auto ), so with 4 modes + auto.

So when CONF_AUTO_FAN_HIGH is setted to high, target_index var = 4 - 2 = 2 (low)

The consequence is that when there is a large T° delta between current value and setpoint, HA tries to set the fan_mode to high, but the underlying device mode is in fact set to low. If i tries to set manually the fan mode high, it falls back to law after a few seconds.

I attach some logs and  yaml state files.
[climate_status.yaml](https://github.com/user-attachments/files/24252948/climate_status.yaml)
[Log_hass.txt](https://github.com/user-attachments/files/24252949/Log_hass.txt)
[vtherm_status.yaml](https://github.com/user-attachments/files/24252950/vtherm_status.yaml)

Didn't found a workaround for that

P.S.: I'm quite new to HA and not a dev,  i speak a little bit python, but I'm not very fluent.

Gamso:
Hello. Indeed when I checked the values for several brands, I didn't see that Hitachi provided a reverse list :(.
We would need to see if we can determine the order of the list, such as searching for the position of “high” or “4,” but I'm afraid it would become too complicated if another brand does it differently.

clockwe:
I've just made a test by adding this condition into the code at line 430:
                                                                                                                                               
        # We suppose speed_modes are ordered from low to high speed                                                                            
        num_speeds = len(speed_modes)                                                                                                          
        if num_speeds == 0:                                                                                                                    
            self._auto_activated_fan_mode = None                                                                                               
            return                                                                                                                             
                                                                                                                                               
        if (speed_modes[0]!="low"):                                                                                                            
            speed_modes.reverse() 

It works more or less...
the calculated fan mode for "CONF_AUTO_FAN_HIGH" is now medium (as index 2 in reversed list), some other modifications are needed, to get the expected fan power, but at least the heating is more efficient.

As you say, I'm also afraid that the code become more complicated, if some manufacturers are using different order or verbs...

Gamso:
This fix is not robust.

For exemple :
- Melcloud return a number list "1,2,3,4,5"
- ProtoART – ClimateControl return "silent, low, med, high, superhigh, auto"

If "auto" is first, it's not a problem because  the code remove “auto” on the list.

clockwe:
I think I found some workarounds without modifying the code, but it will work only for my configuration.

The PAC is driven by the smartIR integration, which sends some IR signals through a Broadlink controller.
This integration provides json files per manufacturers/device (about 380 files) which include device capabilities and IR codes.

For the json file that defines my hvac, I reversed the order of the fan modes from high to diffuse to diffuse to high, so that the order matches the VTherm logic.

The second issue, was that I have 4 modes, but no turbo mode (instead, I have a diffuse mode), so when vtherm configuration is  CONF_AUTO_FAN_HIGH the fan mode switches to medium. I've just changed the auto fan mode to CONF_AUTO_FAN_TURBO on the underlying vtherm configuration.



Regarding your idea of checking fan modes order with key words,  I had a look on SmartIR json files, to see which fashion is used by devices to list the fan modes.

A lot of devices form almost all manufacturers are using a 3 states are using the 3 states low medium high (40 devices) or low-mid-high (121 devices).

Almost all models are using a low to high logic. But there are some exceptions, like my HItachi or some Fujitsu and Toshiba (more than 30 devices), 

Daikin seems to have different approaches to classify fan modes, some are "classical", some are using  a levelX to levelY approach.
Some Daikin devices are using a large number of fan modes (up to 7)
 
As you said, it will quickly become difficult to manage, as keywords name and position for special mode like quiet or turbo not standard at all. Sometimes, quiet mode can be named "breeze", "silent"... and can be placed at the beginning or at the end of the list. Turbo mode are generally at the end of the list, but name can vary ("highest", "powerful"...)

Of course these facts are just regarding SmartIR integrations, Perhaps some other integrations, like your MelCloud or perhaps a Tuya integration, would adopt a different logic.

Hope this will help

wagnst:
@Gamso @jmcollin78 i assume this is not complete. MELCloud home (the new version of MElCloud) emits a different status for the fan speeds. I think it doesn't get recognized correctly. See the entity below:

```
hvac_modes:
  - "off"
  - heat
  - cool
  - dry
  - fan_only
min_temp: 15
max_temp: 30
fan_modes:
  - auto
  - speed_1
  - speed_2
  - speed_3
  - speed_4
swing_modes:
  - auto
  - swing
  - position_1
  - position_2
  - position_3
  - position_4
  - position_5
swing_horizontal_modes:
  - auto
  - swing
  - left
  - left_centre
  - centre
  - right_centre
  - right
current_temperature: 28
temperature: 23
fan_mode: auto
swing_mode: auto
swing_horizontal_mode: centre
friendly_name: AC Bedroom
supported_features: 937
target_temp_step: 1
```

PR #1430, opened 2025-12-20
Feat : Save fan_mode value into StateManager to restore the value when auto fan mode is finished.

Gamso:
I tried to do something, but there is still a problem and some questions about functionality.

My new test failed about a RecursionError :
`FAILED tests/test_auto_fan_mode.py::test_over_climate_auto_fan_mode_turbo_deactivated[True] - RecursionError: maximum recursion depth exceeded`

And other test failed :
`FAILED tests/test_auto_fan_mode.py::test_over_climate_auto_fan_mode_turbo_activation[True] - AssertionError: assert 0 == 1
 +  where 0 = <AsyncMock name='set_fan_mode' id='140020086069904'>.call_count`
 
 For the RecursionError, it' because i use `await self.update_states(force=False)` into `async def async_set_fan_mode(self, fan_mode: str)` function.
 
With my solution, the value of the underlying_entity is no longer taken into account (for ie: user change fan _mode from underlying directly).

I create this PR so you can tell me if my approach is correct and where I'm going wrong.

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -129,7 +129,7 @@ class BaseThermostat(ClimateEntity, RestoreEntity, Generic[T]):
         # self._saved_preset_mode = None
         # self._saved_hvac_mode = None
 
-        self._fan_mode = None
+        # self._fan_mode = None
         self._humidity = None
         self._swing_mode = None
         self._swing_horizontal_mode = None
@@ -175,8 +175,6 @@ class BaseThermostat(ClimateEntity, RestoreEntity, Generic[T]):
         self._minimal_deactivation_delay = None
         self._tpi_threshold_low = None
         self._tpi_threshold_high = None
-
-        self._attr_fan_mode = None
 
         self._is_central_mode = None
         self._last_central_mode = None
@@ -339,7 +337,6 @@ class BaseThermostat(ClimateEntity, RestoreEntity, Generic[T]):
         self._presets_away: dict[str, Any] = {}  # presets_away
 
         self._humidity = None
-        self._fan_mode = None
         self._swing_mode = None
         self._swing_horizontal_mode = None
         self._cur_temp = None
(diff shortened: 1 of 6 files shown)


custom_components/versatile_thermostat/base_thermostat.py
```diff
@@ -176,8 +176,6 @@ def __init__(
         self._tpi_threshold_low = None
         self._tpi_threshold_high = None
 
-        self._attr_fan_mode = None
```
jmcollin78, line 179: _attr_fan_mode is a standard attribute in climate entity. You cannot remove it else it makes Vtherm as non standard climate.

custom_components/versatile_thermostat/state_manager.py
```diff
@@ -79,8 +79,9 @@ async def calculate_current_state(self, vtherm: "BaseThermostat") -> bool:
         change_hvac_mode = await self.calculate_current_hvac_mode(vtherm)
         change_preset = await self.calculate_current_preset(vtherm)
         change_target_temp = await self.calculate_current_target_temperature(vtherm)
+        change_fan_mode = self.update_current_fan_mode_from_requested()
```
jmcollin78, line 82: name it calculate_current_fan_mode as others

custom_components/versatile_thermostat/state_manager.py
```diff
@@ -257,6 +258,11 @@ async def calculate_current_target_temperature(self, vtherm: "BaseThermostat") -
 
         return self._current_state.is_target_temperature_changed
 
+    def update_current_fan_mode_from_requested(self) -> bool:
+        """Update the current fan mode from the given base_thermostat."""
```
jmcollin78, line 262: I should see here: if auto_fan_mode is activated, then current_fan_mode = auto_fan_mode else current_fan_mode =  requested_state.current_fan_mode

custom_components/versatile_thermostat/thermostat_climate.py
```diff
@@ -312,11 +312,9 @@ async def _send_auto_fan_mode(self):
                 self._auto_fan_mode,
                 dtemp,
             )
+            self._auto_deactivated_fan_mode = self._state_manager.current_state.fan_mode
```
jmcollin78, line 315: ok. We need to do a manager for fan_mode. But i guess this is a too complex change.

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/issues/1632

PR #1469, opened 2025-12-28
Speed up tests

Komzpa:
I've noticed that tests often time out when trying to run them in the Codex environment. Here is AI refactored version that is sped up.

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           python-version: "3.13"
       - run: python3 -m pip install "black>=24.0,<25.0"
-      - run: black .
+      - run: black --check .
 
   tests:
     # Tests don't run in Gitlab ci environment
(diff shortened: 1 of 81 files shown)


jmcollin78:
Hello,

I really don't like that. I will need to be able to maintain the code and then to understand what has been done. I try to read but I don't understand what is done.

In VSCode, tests are not so long (3 minutes if I remember well).

Moreover I progressively introduce the `await wait_for_local_condition` instead of a hardcoded sleep.

unit tests is a very sensitive subject: there is more than 350 tests now we many generation of code compatibility. You know when you start a refacto which impacts the tests but you never know when it ends. With my last big refacto (V8 on novembre), i need to fully understand the tests to be able to update them.

jmcollin78:
Unit tests are failed...

jmcollin78:
Hello @Komzpa, I'm very sorry to see that you continue to post changes into this PR. If you read carefully my previous message, I really don't like this PR. 81 files updated and you try to solve an issue that doesn't exists. Tests are not so long and time is acceptable.

I will not apply this PR. Do I need to close it or do you want to do it ?

PR #1475, opened 2025-12-30
Auto-tpi: algorithm improvements post alpha2

### LINKED ISSUE #1451: [Auto-TPI] Learning should account for boiler activation when using central configuration.
Version:  8.4.0.alpha2 

Currently the auto-tpi learning does not account for the activation state of a central boiler which is switched according to VTherm state. Each VTherm is looked at independently, so learning is always attempted when the VTherm indicates that it is heating.

This is problematic with a central boiler configuration because it means an individual VTherm could be calling for heat, but the activation threshold of the boiler is not yet met and so no heat is produced.

If this state persists, en error message is produced and learning is disabled:

> Auto TPI learning for Hall Thermostat has been stopped due to 3 consecutive failures. Reason: Temperature dropped while heating. Please check your configuration.

This is a situation that can arise commonly in a configuration where one VTherm is cooler than others, but the central configuration requires more than one VTherm to be calling for heat before the boiler is started. As that is a normal possible configuration for VTherm, it is not a situation that should cause an error but be handled correctly internally.

Checking for boiler state in the learning validation would avoid this problem. It would also help with the accuracy of the learning by not including cycles where no heat should be expected.

KipK:
This PR should fix & improve first feedbacks from alpha::

- only disable learning when temperature dropped while heating in power saturation
- do not stop learning after consecutive fails in continuous mode
- filter cycles where central boiler is not heating yet ( close #1451 )
- add overshoot compensation mecanism

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_tpi.py
+++ b/custom_components/versatile_thermostat/thermostat_tpi.py
@@ -207,6 +207,19 @@ class ThermostatTPI(BaseThermostat[T], Generic[T]):
             self.hass.async_create_task(self._auto_tpi_manager.async_save_data())
         super().remove_thermostat()
 
+    def _is_central_boiler_off(self) -> bool:
+        """Check if the central boiler is configured but currently off.
+        
+        Returns True if this thermostat is used by a central boiler
+        and that boiler is currently not active.
+        """
+        if not self.is_used_by_central_boiler:
+            return False
+        api = VersatileThermostatAPI.get_vtherm_api()
+        if api and api.central_boiler_manager:
+            return not api.central_boiler_manager.is_on
+        return False
+
     async def _get_tpi_data(self) -> dict[str, Any]:
         """Calculate and return TPI cycle parameters.
         Called by AutoTpiManager at the start of each cycle.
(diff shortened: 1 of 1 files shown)


KipK:
@jmcollin78 I need your review [on this commit, ](https://github.com/jmcollin78/versatile_thermostat/pull/1475/commits/0f35bbb185beb7df64d5386999c1753df7d936ad) , as I really can't test and have a few visibility on how central boiler works. 
I used the VT api for that, can you confirm it was the ok way to do here ? 


When we're good with this PR, you can publish a new test version. I think we should keep alpha state for now until our early testers confirm it solve the first issues we got.

custom_components/versatile_thermostat/thermostat_tpi.py
```diff
@@ -213,12 +214,20 @@ async def _get_tpi_data(self) -> dict[str, Any]:
         # Feed current temperatures to AutoTpiManager BEFORE getting params
         # This ensures the snapshot in on_cycle_started() uses current values
         if self._auto_tpi_manager:
+            # Check central boiler state
```
jmcollin78: Je pense que c'est bien. Par contre, ce bout de code mériterait d'être isolé et réutilisé plus bas.
Et tu pourras le tester tout seul au besoin.

custom_components/versatile_thermostat/thermostat_tpi.py
```diff
@@ -257,12 +266,20 @@ async def _control_heating_specific(self, force=False):
 
         # Feed the Auto TPI manager (Before starting the cycle to apply new coeffs if any)
         if self._auto_tpi_manager:
+            # Check central boiler state
```
jmcollin78: Bout de code à mettre en commun

jmcollin78 (commented):
Toute petite mise en commun et on est bon

custom_components/versatile_thermostat/thermostat_tpi.py
KipK: done with https://github.com/jmcollin78/versatile_thermostat/pull/1475/commits/3895907590d4553d9a1c4001e4ab8f3403bf243a

KipK:
Allez en route to alpha 3 :)

PR #1482, opened 2025-12-30
fix wrong default values from config migration in  __ini__.py in 8.4.0

KipK:
urgent fix to solve migration problem

### DIFF
--- a/custom_components/versatile_thermostat/__init__.py
+++ b/custom_components/versatile_thermostat/__init__.py
@@ -269,19 +269,19 @@ async def async_migrate_entry(hass: HomeAssistant, config_entry: ConfigEntry):
         if version <= 201:
             # Migration 2.1 to 2.2 -> add auto TPI parameters with default values
             new[CONF_AUTO_TPI_MODE] = False
-            new[CONF_AUTO_TPI_ENABLE_UPDATE_CONFIG] = False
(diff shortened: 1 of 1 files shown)


KipK:
But the strange thing is I don't have this migration code on my branch

if version <= 201:
            # Migration 2.1 to 2.2 -> add auto TPI parameters with default values
            ....

KipK:
this commit made this change, not on my side: 

https://github.com/jmcollin78/versatile_thermostat/commit/1d5be07f33aa380b68281ca76bcd453274231388

jmcollin78 (approved):
Thank you !

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/releases/tag/8.4.1

KipK:
It seems it's the only unexpected change that commit https://github.com/jmcollin78/versatile_thermostat/commit/1d5be07f33aa380b68281ca76bcd453274231388 did.
Question is , where did those values and the whole block came from?  I don't like when things appear as dark magic without understanding what happened

jmcollin78:
I remember to merge them, because I change this migration method which was ugly. It comes obviously from you (I'm not able to invent this) but maybe in an old intermediate release. Merge is never is exact science.

KipK:
yeah I've removed the migration part a while ago. I wonder how it could stayed there. Seems all the rest is ok luckily. :)

PR #1484, opened 2025-12-31
fix(auto-tpi): improve overshoot correction

KipK:
improve overshoot correction to avoid false triggers on setpoint change

Before: Overshoot correction triggered whenever temp > setpoint with power > 5%, causing false positives when user lowered the setpoint. 
After: Overshoot correction only triggers when temp is NOT FALLING (stagnating or rising) despite being above setpoint. This correctly handles:
- Setpoint lowered: temp naturally falls → no correction (correct behavior)
- Real Kext error: temp stagnates/rises despite overshoot → correction applied


jmcollin78:
If the PR is not tagged as Draft, I consider I can merge it. Alright ?

KipK:
Yes it's a critical one before we get too much users reporting problems because of the previous overshoot fix. 
( We've talked about it yesterday, that the kind of issue I was a bit afraid to see  ;) )

jmcollin78:
I will wait a few days (-> 02/01 ?) and then do a fix release.

KipK:
> I will wait a few days (-> 02/01 ?) and then do a fix release.

Perfect. I have 2 other new commits improving the algorithm. I'd like them in the next release too, testing it here before PR.

jmcollin78:
I will then wait for you go.

Issue #1458, opened 2025-12-27
Recalling presets is broken since v8.3 for over_climate thermostats (Daikin AC)


soif:
Bonjour.
Tout d'abbord, Félicitations pour ce fantastique composant! Merci pour votre travail !!!
_Lets continue in english_

## Version of the custom_component
- 8.3.2 bugs
- 8.3.1 bugs
- 8.3.0 bugs
- 8.2.0 WORKS !

## Configuration

My VTherm attributes are the following:
```yaml

hvac_modes:
  - fan_only
  - dry
  - cool
  - heat
  - "off"
min_temp: 7
max_temp: 35
target_temp_step: 0.5
fan_modes:
  - Auto
  - Silence
  - "1"
  - "2"
  - "3"
  - "4"
  - "5"
preset_modes:
  - none
  - frost
  - eco
  - comfort
  - boost
swing_modes:
  - "Off"
  - Vertical
  - Horizontal
  - 3D
current_temperature: 18.5
temperature: 23.5
fan_mode: "4"
hvac_action: heating
preset_mode: none
swing_mode: Vertical
hvac_mode: heat
ema_temp: 18.12
specific_states:
  is_on: true
  last_central_mode: null
  last_update_datetime: "2025-12-27T10:58:31.693921+01:00"
  ext_current_temperature: 9.65629196166992
  last_temperature_datetime: "2025-12-27T10:57:50.469171+01:00"
  last_ext_temperature_datetime: "2025-12-27T10:58:31.693252+01:00"
  is_device_active: true
  device_actives:
    - climate.sejour
  nb_device_actives: 1
  ema_temp: 18.12
  temperature_slope: 3.35
  hvac_off_reason: null
  total_energy: 0
  last_change_time_from_vtherm: "2025-12-27T10:48:11.793260+01:00"
  messages: []
  is_sleeping: false
  is_locked: false
  is_recalculate_scheduled: true
configuration:
  ac_mode: true
  type: over_climate
  is_controlled_by_central_mode: false
  target_temperature_step: 0.5
  minimal_activation_delay_sec: 10
  minimal_deactivation_delay_sec: 0
  timezone: Europe/Paris
  temperature_unit: °C
  is_used_by_central_boiler: false
  max_on_percent: null
  have_valve_regulation: false
  cycle_min: 5
preset_temperatures:
  frost_temp: 9
  eco_temp: 15
  boost_temp: 22
  comfort_temp: 17
[...]

PR #1496, opened 2025-12-31
Fix # 1458 - Some AC units (e.g. Daikin) do not handle multiple consecutive commands well

Gamso:
Two mechanisms have been introduced to prevent this :
- a locking mechanism when multiple commands compete.
- a throttling mechanism that adds a 1-second delay when commands are sent too close together.

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -1,6 +1,7 @@
 # pylint: disable=line-too-long, too-many-lines, abstract-method
 """ A climate over climate classe """
 import logging
+
 from datetime import timedelta, datetime
 
 from homeassistant.const import STATE_UNAVAILABLE, STATE_UNKNOWN
--- a/custom_components/versatile_thermostat/underlyings.py
+++ b/custom_components/versatile_thermostat/underlyings.py
@@ -3,8 +3,9 @@
 """ Underlying entities classes """
 import logging
 import re
-from typing import Any, Dict, List, Tuple, TypeVar
+from typing import Any, Dict, List, Optional, Tuple, TypeVar
 from collections.abc import Callable
+from datetime import datetime, timedelta
 
 from enum import StrEnum
 
@@ -13,7 +14,7 @@ from homeassistant.core import State
 
 from homeassistant.exceptions import ServiceNotFound
 
-from homeassistant.core import HomeAssistant, CALLBACK_TYPE
+from homeassistant.core import HomeAssistant, CALLBACK_TYPE, Context, ServiceResponse
 from homeassistant.components.climate import (
     ClimateEntity,
     ClimateEntityFeature,
@@ -81,6 +82,7 @@ class UnderlyingEntity:
         self._entity_id: str = entity_id
         self._hvac_mode: VThermHvacMode | None = None
         self._on_cycle_start_callbacks: list[Callable] = []
+        self._last_command_sent_datetime: datetime = datetime.fromtimestamp(0)
 
     def register_cycle_callback(self, on_start: Callable):
         """Register a callback for cycle start"""
@@ -178,6 +180,22 @@ class UnderlyingEntity:
         """Call the method after a delay"""
         return async_call_later(hass, delay_sec, called_method)
 
+    async def hass_services_async_call(
+        self,
+        domain: str,
+        service: str,
+        service_data: dict[str, Any] | None = None,
+        blocking: bool = False,
+        context: Context | None = None,
+        target: dict[str, Any] | None = None,
+        return_response: bool = False,
+    ) -> ServiceResponse:
+        """Wrapper for HASS service calls"""
+        reponse: ServiceResponse = await self._hass.services.async_call(domain, service, service_data, blocking, context, target, return_response)
+
+        self._last_command_sent_datetime = self._thermostat.now
+        return reponse
+
     async def start_cycle(
         self,
         hvac_mode: VThermHvacMode,
@@ -218,7 +236,7 @@ class UnderlyingEntity:
             data = {"value": value}
             target = {ATTR_ENTITY_ID: number_entity_id}
             domain = number_entity_id.split(".")[0]
-            await self._hass.services.async_call(
+            await self.hass_services_async_call(
                 domain=domain,
                 service=SERVICE_SET_VALUE,
                 service_data=data,
@@ -600,8 +618,9 @@ class UnderlyingClimate(UnderlyingEntity):
             entity_type=UnderlyingEntityType.CLIMATE,
             entity_id=climate_entity_id,
         )
-        self._underlying_climate = None
-        self._last_sent_temperature = None
+        self._underlying_climate: Optional[ClimateEntity] = None
+        self._last_sent_temperature: Optional[float] = None
+        self._cancel_set_fan_mode_later: Optional[Callable[[], None]] = None
 
     def find_underlying_climate(self) -> ClimateEntity:
         """Find the underlying climate entity"""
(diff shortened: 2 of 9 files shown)


Gamso:
The throttling mechanism will likely increase the test duration. I’ll let you decide whether this is acceptable.
We may also need to evaluate whether a 1-second delay is sufficient.

custom_components/versatile_thermostat/thermostat_climate.py
```diff
@@ -121,6 +129,36 @@ def post_init(self, config_entry: ConfigData):
...
+                elapsed = now - last
+                if elapsed < self._cmd_delay:
+                    delay = self._cmd_delay - elapsed
+                    _LOGGER.debug(
+                        "%s - #1458 - Delaying command for underlying %s by %.2fs",
+                        self,
+                        under.entity_id,
+                        delay,
+                    )
+                    await asyncio.sleep(delay)
```
jmcollin78: I really think this is bad design to have some asyncio.sleep in core feature even for one second. When I need this (and I already do this), then I use the async_call_later which does a call later in time. You already have example in ThermostatClimate.do_send_regulated_temp_later. Which send the regulation temp later in time if one has already been sent recently.

custom_components/versatile_thermostat/thermostat_climate.py
```diff
@@ -1235,7 +1272,7 @@ def turn_aux_heat_on(self) -> None:
     async def async_turn_aux_heat_on(self) -> None:
         """Turn auxiliary heater on."""
         for under in self._underlyings:
-            await under.async_turn_aux_heat_on()
+            await self._execute_underlying_command(under, under.async_turn_aux_heat_on())
```
jmcollin78: I think, it is preferable to delegate this delay check directly in the UnderlyingClimate class. Then it is transparent for all ThermostatClimate and ThermostatClimateValve class. All the method `set_xxx` are defined in UnderlyingClimate and could be used.

The other thinking is that before your PR, we never had this issue. So maybe we could just implement the delay in the method `set_fan_mode`. The others seem to not have any issue.

I'm afraid of side effect to delay all potential commands. There are so many combinations possible in `over_climate` with so many different brands and devices, I don't want to take a little risk on that. 

What will happen when I change the hvac_mode and the setpoint and the fan_mode at the same (which is possible if I just set hvac_mode to heat) ?   Are the 3 commands delayed ? In what order ? If you delay setpoint and fan_mode they will be send at the same time not ? Then to be totally exhaustive we will need to implement a queue with all commands waiting and take them second by second to send them.  Very complicated and very risky for the brands and device diversity. I really prefer to delay only the `set_fan_mode` command to avoid side effects.

What do you think ?

jmcollin78:
Another question sorry. Do you have a Daikin AC unit ? Are you able to see it is fix the case or do we need to ask to the user ?

Gamso:
We need to ask to the user. I don't know if 1 second is enough.

custom_components/versatile_thermostat/thermostat_climate.py
Gamso: I agree. I just tried a more generic approach for potential multiple command issues, but just on the `set_fan_mode ` should be fine.
In fact, I saw that the problem was in `async def async_control_heating(self, timestamp=None, force=False) -> bool:`. Maybe just add delay into this function ?

jmcollin78: `async_control_heating` is badly named. It is the entry point of the monitoring cycle. So when there is something to do, all commands are triggered by this function. You cannot delay it.

jmcollin78:
> We need to ask to the user. I don't know if 1 second is enough.

ok. So when an acceptable fix will be ready, I will publish a beta to this specific user #1458

custom_components/versatile_thermostat/thermostat_climate.py
Gamso: Do you want me to keep the command memory to add a delay to `set_fan_mode` only if another command is sent, or should I add a delay every time?

jmcollin78: Humm, good question. After many though here is my definitive proposition:
1. In UnderlyingClimate only,
2. Each time you have a `await self._hass.services.async_call` then you store the `self._thermostat.now` datetime into a local `last_command_sent_datetime`
3. Only in the `set_fan_mode` method:
- you check if `self._thermostat.now` is > `last_command_sent_datetime` + 2 sec
- if ok you send the command and set `last_command_sent_datetime`
- else you call back the `set_fan_mode` method 2 sec later (async_call_later)

Then we will have a robust and simple mecanism for delaying command that is used only for fan mode. This mecanism will be usable for the next time we have such issues.

I propose 2 sec to more sure the device has time to ingest the previous command, but this is an empiric magic number coming from nowhere. I don't want to make it configurable. With that mecanism we could also set this to 5 sec easily.

Thank you for your work !

Gamso: Do you have a tips to test in debug `UnderlyingClimate.set_fan_mode()` ?
 Because i try on several tests but mockEntity bypass this function :(.

Gamso: ok I find a solution to test thanks to Gemini : 

> original_method = UnderlyingClimate.set_fan_mode
> 
> with patch(
>     "custom_components.versatile_thermostat.underlyings.UnderlyingClimate.set_fan_mode",
>     autospec=True,
>     side_effect=original_method # Ceci force l'exécution du vrai code
> ) as mock_send_fan_mode:

custom_components/versatile_thermostat/thermostat_climate.py
Gamso: reverted !

jmcollin78 (commented):
Good for me until now. Just a test is missing I guess: a test with a previous call, and a test without is sufficient I guess.

jmcollin78:
Some test failed.

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -637,6 +640,22 @@ def is_initialized(self) -> bool:
         """True if the underlying climate was found"""
         return self._underlying_climate is not None
 
+    async def hass_services_async_call(
```
jmcollin78: Put this into common class UnderlyingEntity (at top of the file) and it should be ok for the tests

jmcollin78:
This change has break many tests. Do you need help ?

Gamso:
Finally all green !
I add test into tests/test_auto_fan_mode.py to check delay

custom_components/versatile_thermostat/underlyings.py
```diff
@@ -178,6 +180,22 @@ def call_later(
...
+        domain: str,
+        service: str,
+        service_data: dict[str, Any] | None = None,
+        blocking: bool = False,
+        context: Context | None = None,
+        target: dict[str, Any] | None = None,
+        return_response: bool = False,
+    ) -> ServiceResponse:
+        """Wrapper for HASS service calls"""
+        reponse: ServiceResponse = await self._hass.services.async_call(domain, service, service_data, blocking, context, target, return_response)
```
jmcollin78, line 194: If you have keep the previous form:
```
await self._hass.services.async_call(
                domain=domain,
                service=SERVICE_SET_VALUE,
                service_data=data,
...
```

Then you don't need to change all the tests.  But that is ok, I should look at this before.

jmcollin78 (approved):
Approved ! You can merge it.

custom_components/versatile_thermostat/underlyings.py
Gamso: Actually, I saw one call with 4 parameters and the others with 3. I should have gone back over the tests anyway. At least now we have a wrapper that doesn’t distort the possible parameters.

jmcollin78: ok fine. Let's merge it.

jmcollin78:
Thank you for your awesome work !

PR #1503, opened 2026-01-01
Auto tpi domains

KipK:
Introduce KEXT_LEARNING_MAX_GAP (0.5°C) to block Kext learning during far-field stagnation.
- Prevents `Kext` from being incorrectly inflated during transient ramp-up phases, reducing overshoot.
- Stagnation > 0.5°C is now treated correctly as a `Kint` or Capacity issue.
- Enable kint boost by default


KipK:
we will be able to enable Kint boost with this now. 
Following [discussion](https://github.com/jmcollin78/versatile_thermostat/discussions/1428#discussioncomment-15387286)

PR #1511, opened 2026-01-02
Refactor/typing pylance cleanup

Gamso:
While browsing the codebase, I noticed many warnings reported by Pylance.
I reduced them where it could be done safely.

I split the changes into multiple commits to keep things clearer.

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -5,10 +5,9 @@
 import math
 import logging
 import asyncio
-from datetime import datetime, timedelta
-from functools import partial
+from datetime import datetime
 
-from homeassistant.components.recorder import history, get_instance
+from homeassistant.components.recorder import history
 from homeassistant.util import dt as dt_util
 from typing import Any, Generic
 from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


custom_components/versatile_thermostat/ema.py
```diff
@@ -26,13 +26,13 @@ def __init__(
         vterm_name: str,
         halflife: float,
         timezone: tzinfo,
-        precision: int = 3,
-        max_alpha: float = 0.5,
+        precision: int,
```
jmcollin78, line 29: You remove the default value ?

custom_components/versatile_thermostat/auto_start_stop_algorithm.py
```diff
@@ -55,10 +55,10 @@
 class AutoStartStopDetectionAlgorithm:
     """The class that implements the algorithm listed above"""
 
-    _dt: float | None = None
-    _level: str = AUTO_START_STOP_LEVEL_NONE
-    _accumulated_error: float = 0
-    _error_threshold: float | None = None
+    _dt: float = 0.0
```
jmcollin78, line 58: Changing typing and default values is not safe. I will not take this PR into account. Too dangerous.

jmcollin78:
Sorry but I will not take this PR.

custom_components/versatile_thermostat/ema.py
Gamso: Yes, because it is never used in the code: all values are always passed via parameters. Only max_alpha is used in a test.

custom_components/versatile_thermostat/auto_start_stop_algorithm.py
Gamso: I change `_level` type because in the code you always set a value from TYPE_AUTO_START_STOP_LEVELS.

jmcollin78:
Sorry but i will not take this.

Issue #1144, opened 2025-10-24
[Feature Request]: Heating Anomaly Detection (Valve 100% but no temp rise)


nicola-spreafico:
Hi @jmcollin78 (and community),

First, thank you for this fantastic integration! I'm using Versatile Thermostat to control several Sonoff TRV Zigbee valves, and it works great.

I'm looking for a way to monitor the health of my heating system, and I'd like to propose a new feature, or ask if this is already possible.

**The Problem / Use Case**

This request comes from a real failure I experienced just today.

One of my rooms failed to heat up. When I checked, the Versatile Thermostat was correctly in heat mode and the TRV valve_position had been at 100% for over 18 hours.

The problem was a stuck electrovalve (zone valve) for that heating loop, so no hot water was circulating. Nobody noticed the problem until the room was freezing, and the system wasted energy trying to heat an unresponsive zone.

This highlights a critical monitoring gap. The general failure scenario I want to detect is:
- The Versatile Thermostat is calling for heat (state is heat).
- The TRV's valve_position (which this integration exposes) goes to 100% (or > 95%).
- The valve stays at 100% for an extended period (e.g., 30-45 minutes).
- However, the current_temperature does not increase and remains significantly below the target temperature (setpoint).

This situation clearly indicates a problem (e.g., the central boiler is off, air is trapped in the radiator, the TRV is stuck, or as in my case, a stuck zone valve).

**My Question / Feature Proposal**

1. Does a native feature for this already exist? Is there already a built-in sensor or attribute in Versatile Thermostat that detects this kind of anomaly? I've checked the documentation but couldn't find one.

2. If not, could this be added as a native feature? It would be incredibly useful if the Versatile Thermostat could optionally expose a binary_sensor for this.

For example: binary_sensor.my_thermostat_heating_anomaly

This sensor would turn to on only when the "anomaly" condition is met (valve 100% for X minutes, but temperature is not rising or is still far from the setpoint).

This would allow users to easily create automations for critical notifications (e.g., "Warning: The living room is not heating up despite the valve being fully open!") without everyone needing to build their own complex template automations.

What are your thoughts? Is this feasible, or perhaps there's a simpler way to achieve this that I've missed?

Thanks for your consideration!

jmcollin78:
Hello @nicola-spreafico ,

> Does a native feature for this already exist?

[...]

PR #1525, opened 2026-01-04
Feat: #1144 - Detection of heating anomaly

jmcollin78:


bdemaistre:
Thanks for this Version : 8.5.0.beta2. I will try with some error case but I have 2 questions 

1. How the tolerance is affecting the calculation ? With 15 min and 0,5°C tolerance it means that the error is triggered when the rise is less than 0.5°C in 50min ? (this is quite a strong rise already, and also the thermometers seems to be a lot more accurate for following temperature at 0.1 or 0.2 °C change than for determining absolute temperature at 0.5 or 1° precision)
2. I know that I am quite bad at copying and pasting Yaml code but even changing identation following marked errors I could not use the proposed automation code (in documentation/fr/feature-heating-failure-detection.md). I had to create simple automation in visual editor for each vtherm (using "Déclencheur" "thermostat XXX État d'anomalie de chauffe a commencé à détecter un problème"). Is this code ready for copy paste or it needs some modification ?

jmcollin78:
Hello thank you @bdemaistre for this comment. You should have post this into the [#1144 ticket](https://github.com/jmcollin78/versatile_thermostat/issues/1144)  because this will disapear after merging, but anyway here are my answers:

> Thanks for this Version : 8.5.0.beta2. I will try with some error case but I have 2 questions
> 
> 1. How the tolerance is affecting the calculation ? With 15 min and 0,5°C tolerance it means that the error is triggered when the rise is less than 0.5°C in 50min ? (this is quite a strong rise already, and also the thermometers seems to be a lot more accurate for following temperature at 0.1 or 0.2 °C change than for determining absolute temperature at 0.5 or 1° precision)

not 0,5°C in 50 min but in 15 min. Yes the temp should rise at least for 0,5° in 15 min. All this parameters are tunable to fit your installation.

> 2. I know that I am quite bad at copying and pasting Yaml code but even changing identation following marked errors I could not use the proposed automation code (in documentation/fr/feature-heating-failure-detection.md). I had to create simple automation in visual editor for each vtherm (using "Déclencheur" "thermostat XXX État d'anomalie de chauffe a commencé à détecter un problème"). Is this code ready for copy paste or it needs some modification ?

I detect some indentation bugs but I fixed it yesterday afternoon. It should be good to implement the automations. That is what I have posted here: https://github.com/jmcollin78/versatile_thermostat/issues/1144

PR #1534, opened 2026-01-05
feat: Add startup sync to re-evaluate conditions after HA restart

Le-Syl21:
This feature addresses the issue where VTherm restores the previous state after a Home Assistant restart without re-evaluating current conditions. For example, if presence changed while HA was down, VTherm would not detect this until manual intervention.

Changes:
- Add new configuration options in the Advanced section:
  - startup_sync_enabled: Enable/disable the feature (default: true)
  - startup_sync_max_wait_sec: Max time to wait for temp sensor (default: 300s)
  - startup_sync_safety_delay_sec: Safety delay after sensor available (default: 10s)

- Implement _async_startup_sync() method that:
  - Waits for the temperature sensor to become available (important for Zigbee/WiFi devices that take time to initialize after HA restart)
  - Adds a configurable safety delay
  - Re-evaluates all feature manager states (presence, window, motion, etc.)
  - Fires a STARTUP_SYNC_EVENT if state changes

- Add translations for EN and FR

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -184,6 +184,12 @@ class BaseThermostat(ClimateEntity, RestoreEntity, Generic[T]):
         self._last_central_mode = None
         self._is_used_by_central_boiler = False
 
+        # Startup sync configuration
+        self._startup_sync_enabled = None
+        self._startup_sync_max_wait_sec = None
+        self._startup_sync_safety_delay_sec = None
+        self._startup_sync_done = False
+
         self._support_flags = None
         # Preset will be initialized from Number entities
         self._presets: dict[str, Any] = {}  # presets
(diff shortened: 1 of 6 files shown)


jmcollin78:
Hello,


I don't know what your are trying to fix. I'm not sure there is an issue with that. Because sensors are listened and new state is then taken into account when they comes to life first.

I have never experienced any issue like that for more than 2 years now.

What is your case ?

> For example, if presence changed while HA was down, VTherm would not detect this until manual intervention.

I think this assumption is false. Each manager reevaluate its state at startup. It should works without needed a new option parameter.

I really prefer to have a real analysed issue before proposing some code changes done with Claude and then not understood.

Le-Syl21:
Bonjour,

Non mais c'est de la merde oublie.

Je vais plutôt me manger la doc

Sylvain GARGASSON
+33 (0)6 63 89 48 80


Le lun. 5 janv. 2026 à 14:56, Jean-Marc Collin ***@***.***> a
écrit :

jmcollin78:
Attention au vibe coding. C'est très chouette et Claude est très fort mais le tout n'est pas de générer du code mais de le maintenir ensuite. Et c'est moi le mainteneur. Si tu lui demandes de résoudre un soucis qui n'existe pas, il va le faire qad même.

Je te laisse la clore.

PR #1553, opened 2026-01-06
feat: pause Auto TPI learning on heating failure detection

KipK:
Use new heating failure feature to pause/start AutoTPI.


jmcollin78 (approved):
Excellente idée !

PR #1561, opened 2026-01-07
auto-TPI: better directional protection robustness

KipK:
Only activate directional protection when power is <5%


KipK:
I'm closing it. I need to do more test on this to compare. I feel it's not necessary.

PR #1566, opened 2026-01-07
Fix Error Entity sensor.total_power_active_for_boiler (<class 'custom…

kvanbiesen:
…_components.versatile_thermostat.sensor.TotalPowerActiveDeviceForBoilerSensor'>) is using native unit.....

Logger: homeassistant.components.sensor
Source: components/sensor/__init__.py:729
integration: Sensor (documentation, issues)
First occurred: 03:02:58 (1 occurrence)
Last logged: 03:02:58

Entity sensor.total_power_active_for_boiler (<class 'custom_components.versatile_thermostat.sensor.TotalPowerActiveDeviceForBoilerSensor'>) is using native unit of measurement 'None' which is not a valid unit for the device class ('power') it is using; expected one of ['GW', 'TW', 'W', 'MW', 'kW', 'mW']; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/jmcollin78/versatile_thermostat/issues

When you only have TRV's in the house, the power sensor just counts those, so it throws a warning on startup as energy sensors need a unit, so then it should be just a DIAGNOSTIC sensor

### DIFF
--- a/custom_components/versatile_thermostat/feature_central_boiler_manager.py
+++ b/custom_components/versatile_thermostat/feature_central_boiler_manager.py
@@ -6,6 +6,7 @@ from datetime import timedelta
 from homeassistant.core import (
     HomeAssistant,
     HomeAssistantError,
+    CoreState,
 )
 
 from homeassistant.helpers.entity import Entity
@@ -98,6 +99,16 @@ class FeatureCentralBoilerManager(BaseFeatureManager):
         controls this central boiler"""
 
         _LOGGER.debug("%s - calculating the new central boiler state", self)
+
+        # Skip service calls if HA is not fully running (during startup/restore)
+        # This prevents calling services before underlying entities are available
+        if self._hass.state != CoreState.running:
+            _LOGGER.debug(
+                "%s - HA not fully started yet (state=%s). Skipping central boiler service calls.",
+                self,
+                self._hass.state,
+            )
+            return False
 
         def _send_vtherm_event(data: dict):
             send_vtherm_event(


kvanbiesen:
Also fixed a racing issue with the boiler restore setting, now it wait until the integration is loaded before restoring the boiler. The switch might not be loaded yet 

Say versatile loads before Shelly or esphome integration. IT throws a warning

custom_components/versatile_thermostat/feature_central_boiler_manager.py
```diff
@@ -99,6 +100,16 @@ async def calculate_central_boiler_state(self, _):
 
         _LOGGER.debug("%s - calculating the new central boiler state", self)
 
+        # Skip service calls if HA is not fully running (during startup/restore)
+        # This prevents calling services before underlying entities are available
+        if self._hass.state != CoreState.running:
```
jmcollin78, line 105: Should be already managed by the is_ready flag. see below. If not ready, then return. Line 136.
And it should be ready when all entities are discovered. So I'm not sure this is useful.

custom_components/versatile_thermostat/sensor.py
```diff
@@ -160,13 +161,23 @@ async def async_my_climate_changed(self, event: Event = None):
...
+    
+    @property
+    def entity_category(self):
+        if not self.my_climate:
+            return EntityCategory.DIAGNOSTIC
+        return None
 
     @property
     def device_class(self) -> SensorDeviceClass | None:
+        if not self.my_climate:
```
jmcollin78, line 173: I have to look more at this. I'm not sure changing device_class is accepted after creating the entity. I never seen that elsewhere (but I don't have seen so much of course). Have you reference or example to this kind of pattern ?

jmcollin78 (commented):
I'm not sure of your PR. I need more test. The initial message is only a matter of unit. Because we have an energy sensor which doesn't have unit (wh, kwh, ....). This make sense in Vtherm because unit is not important. All we need is that all power and energy have the same unit. But HA display a warning (and just a warning which have no impact for me).

custom_components/versatile_thermostat/feature_central_boiler_manager.py
kvanbiesen: I can guarantee you it doesnt, It throws a warning (once at startup) then it works.  My boiler switch is connected with wifi (ESPHOME) and its the first warning that get throw. Disabling the integration removes the warning, 

I tried replacing it with a another test switch that connects tru the shelly integration, and that doesnt throw an error. So those wait for eachother, only problem is it seems with the ESPHOMe flashed switch. Adding that line removes the warning and doesnt harm anything.

custom_components/versatile_thermostat/sensor.py
kvanbiesen: It seems to work for me, no errors thrown, ofcourse i already have a running Vtherm setup so i dont kow what happens. 

Its either that or use a seperate sensor cause puttting the TRV count in a energy sensor is not the best way of storing that. 

Also the unit changes on the fly, is set to WATT_PER_HOUR which is also uncommon. It show always be KW/Hour an let the user set in the gui what he prefers :) Return None for a energy unit, is not commonly done. Hence it also throws an warning

jmcollin78:
I keep that for a next release. I have urgent fix to  release and I don't want to take a risk with your changes.

kvanbiesen:
> I keep that for a next release. I have urgent fix to release and I don't want to take a risk with your changes.

entity sensor Il sfill Looking into. The sensor not ready is defently not needed iin some edge case where the while system is ready . It's only a startup error, I know and it works. But better. No errors that cause troubles, or delays. (the the core state ready one)

I do have to look into the other sensor little more cause not saying its bad coding, its using a non intended cause for ta energy sensor. Shouldnt throw warnings imho if possible unless something is actually wrong.
[8d41ae5](https://github.com/jmcollin78/versatile_thermostat/pull/1566/commits/8d41ae5150db55818b9abcd83bff480f02d9543f) is defently worth merging. 

[887c023](https://github.com/jmcollin78/versatile_thermostat/pull/1566/commits/887c023a380e8109748e7c2b3c87cf0eb4c7538f) i have to test some more, it doesnt break anything for sure, cause altho it works, once every 20 reboots it still shows up occasionally (im also dev of a integration so i also restart a lot :) )

BTw you can change classes and categories on the fly, even after creating. Sure it could drop out of energy board but that shoulndt cause much harm. Maybe its better to add like a counter attribute to the sensor, if if actually counts devices in stead of power, then change all the category/class/type . If it doesnt act as a counter, then use your system.

custom_components/versatile_thermostat/feature_central_boiler_manager.py
jmcollin78: This should be totally fixed with the next major release 9.x. I change all the code around this. So I will not take this into account.

custom_components/versatile_thermostat/sensor.py
jmcollin78: I will take this for 9.x. Thank you.

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/pull/1738

PR #1579, opened 2026-01-09
fix(auto_tpi): reset current_cycle_params on learning start

KipK:
When starting Auto TPI learning with reset_data=True, current_cycle_params was not being reset to None. This caused the first tick to incorrectly try to complete a cycle that never started


KipK:
celui la si t'as pas encore release tu pourra le mettre dedans, il m'avait échappé.

PR #1600, opened 2026-01-11
fix(auto-tpi): use sample-and-hold interpolation for calibrate_capacity service

KipK:
Replace tolerance-based power interpolation with sample-and-hold logic in calculate_capacity_from_slope_sensor.
This fixes an issue where power_percent would produce many "invalid" data points when power stays stable for long periods. 

The new _get_power_at_time_sample_hold method returns the last known power value before the target timestamp, which correctly handles sensors that only report on value change.

It gives a lot more data points available now.

Fixes: power interpolation returning None during stable 100% power periods
fix: #1592


KipK:
rebased

PR #1609, opened 2026-01-12
Refactor Auto TPI configuration

KipK:
- Introduce "Discovery" and "Fine Tuning" presets for simpler setup.
- Add "Enable Advanced Settings" to hide expert parameters.
- Hardcode standard options (update config, notifications, etc.) to simplify UI.
- remove old keys in migration

image


jmcollin78 (approved):
Encore une belle PR !

PR #1616, opened 2026-01-13
feat(auto-tpi): retroactively adjust capacity on Kext updates

KipK:
- Ensure the stored adiabatic capacity remains consistent with the heat loss model by retroactively adjusting it whenever the outdoor coefficient (Kext) is updated. This prevents divergence in the thermal model as Kext evolves.
- [Refactor capacity adjustment to use shared adiabatic formula](https://github.com/jmcollin78/versatile_thermostat/commit/67296cfa19166f8c5d271d15891b631214850d5b)


KipK:
pour faire simple, quand on update Kext, il faut aussi ajuster la capacité adiabatique calculée pour en tenir compte.

jmcollin78 (approved):

PR #1638, opened 2026-01-17
Fix: Auto TPI forces reset when starting a new session

KipK:
Fix: Auto TPI forces reset when starting a new session

Reset of learning data when starting a new session to force using configured coefficient and not remaining one in the TPI loop. Added regression test: test_initialization_reset_bug.


jmcollin78 (approved):

jmcollin78:
I break the test with my last change. I will have to fix that.

Issue #1660, opened 2026-01-21
Max open valve is limited to 100, but Eurotronics Comet uses 255 steps for valve opening degree


mrs200165:
## Version of the custom_component
8.6.0

## Configuration

```
hvac_modes: heat, sleep, off
min_temp: 3
max_temp: 24
target_temp_step: 0.1
preset_modes: none, frost, eco, comfort, boost
current_temperature: 20.9
temperature: 21
hvac_action: heating
preset_mode: none
hvac_mode: auto
ema_temp: 20.93
specific_states: 
is_on: true
last_central_mode: null
last_update_datetime: '2026-01-21T13:51:19.695009+01:00'
ext_current_temperature: 0.6
last_temperature_datetime: '2026-01-21T13:49:09.154075+01:00'
last_ext_temperature_datetime: '2026-01-21T13:49:50.928594+01:00'
is_device_active: true
device_actives:
  - number.comet_trv_arbeitszimmer_valve_position
nb_device_actives: 1
ema_temp: 20.93
temperature_slope: -0.36
hvac_off_reason: null
total_energy: 0.07
last_change_time_from_vtherm: '2026-01-21T13:51:19.694813+01:00'
messages: []
is_sleeping: false
is_locked: false
is_recalculate_scheduled: false
auto_tpi_state: 'off'
auto_tpi_learning: {}

configuration: 
ac_mode: false
type: over_climate_valve
is_controlled_by_central_mode: false
target_temperature_step: 0.1
timezone: Europe/Berlin
temperature_unit: °C
is_used_by_central_boiler: false
max_on_percent: null
have_valve_regulation: true
cycle_min: 1
minimal_activation_delay_sec: 10
minimal_deactivation_delay_sec: 0

preset_temperatures: 
frost_temp: 3
eco_temp: 3
boost_temp: 3
comfort_temp: 3
frost_away_temp: 0
eco_away_temp: 0
boost_away_temp: 0
comfort_away_temp: 0

current_state: 
hvac_mode: auto
target_temperature: 21
preset: none

requested_state: 
hvac_mode: auto
target_temperature: 21
preset: none

is_presence_configured: false
is_power_configured: false
power_manager: 
device_power: 1
mean_cycle_power: 0.18

is_motion_configured: false
is_window_configured: true
is_window_auto_configured: false
window_manager: 
window_state: 'off'
window_auto_state: unavailable
window_action: window_frost_temp
is_window_bypass: false
[...]

PR #1676, opened 2026-01-23
Fix calibration service for valve

### LINKED ISSUE #1656: [Feature Request] - Add a recalibration service (action)
This service aims to fix recalibration issue encountered by some users with their Sonoff TRVZB.

Some users reports that calibration issues with Sonoff TRVZB can be fixed by this scenario:
1. fully open the valve (100%),
2. wait a few,
3. fully close the valve (0%).

This feature should not conform to the 4 parameters which controls the valve openess (opening_threshold, min_opening_degree, max_opening_degree and max_closing_degree).

This aims to fix the issues of this subject: https://github.com/jmcollin78/versatile_thermostat/discussions/1369, https://github.com/jmcollin78/versatile_thermostat/discussions/1429, (and more).

Please add :+1 on this FR if you are interested.

bontiv:
Fix #1656

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate_valve.py
+++ b/custom_components/versatile_thermostat/thermostat_climate_valve.py
@@ -100,6 +100,7 @@ class ThermostatOverClimateValve(ThermostatProp[UnderlyingClimate], ThermostatOv
             opening = opening_list[idx]
             closing = closing_list[idx] if idx < len(closing_list) else None
             self._opening_threshold_degree = max(self._opening_threshold_degree, regulation_threshold)
+            opening_entity = self._hass.states.get(opening)
 
             under = UnderlyingValveRegulation(
                 hass=self._hass,
@@ -108,11 +109,14 @@ class ThermostatOverClimateValve(ThermostatProp[UnderlyingClimate], ThermostatOv
                 closing_degree_entity_id=closing,
                 climate_underlying=self._underlyings[idx],
                 min_opening_degree=(min_opening_degrees_list[idx] if idx < len(min_opening_degrees_list) else 0),
-                max_opening_degree=(max_opening_degrees_list[idx] if idx < len(max_opening_degrees_list) else 100),
+                max_opening_degree=(max_opening_degrees_list[idx] if idx < len(max_opening_degrees_list) else opening_entity.attributes.get("max", 100) if opening_entity else 100),
                 max_closing_degree=self._max_closing_degree,
                 opening_threshold=self._opening_threshold_degree,
             )
             self._underlyings_valve_regulation.append(under)
+
+        # Guard to prevent concurrent recalibration tasks per thermostat entity
+        self._recalibrate_lock: asyncio.Lock | None = None
 
     @overrides
     def restore_specific_previous_state(self, old_state: State):
--- a/.devcontainer/lovelace
+++ b/.devcontainer/lovelace
@@ -1042,7 +1042,9 @@
                   "disable_auto_fan_infos": false,
                   "disable_autoStartStop": false,
                   "disable_circle": true,
-                  "autoStartStopEnableEntity": "switch.thermostat_climate_2_enable_auto_start_stop"
+                  "autoStartStopEnableEntity": "switch.thermostat_climate_2_enable_auto_start_stop",
+                  "theme": "gunmalmg",
+                  "name": "Parents bedroom thermostat"
                 },
                 {
                   "type": "thermostat",
@@ -2597,7 +2599,10 @@
                   "disable_cool": true,
                   "disable_auto": true,
                   "disable_sleep": false,
-                  "set_current_as_main": true
+                  "set_current_as_main": true,
+                  "theme": "gunmalmg",
+                  "allow_lock_toggle": true,
+                  "lock_relock_delay": 10
                 },
                 {
                   "type": "vertical-stack",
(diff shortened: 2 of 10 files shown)


jmcollin78:
Hello, thanks for this PR.

The issue in #1656  is that when we send hvac_mode off to VTherm, it turns off the TRV also. And when the TRV is off, the number change has no effect. With Sonoff TRVZB that is what I see. If I set hvac_mode to heat at the beginiing of the recalibration process, it works. To fix this, we must turn on the TRV after turning off VTherm and before begining the full on/off sequence. This is not so easy because, if the TRV changes, VTherm will listen to this change and update itself.

For me this has nothing to do with the lock.
I understand what you have done and it should be fine - mainly it prevent a hvac_mode change during the recalibration - but doesn't fix the main issue for me. But what you have done will allow us to detect that a recalibration is running and then avoid changing the VTherm hvac_mode when the TRV state change.

Other thing, you have set the closing degree number entity as optionnal. But the principle of recalibration is to do a full open and close. Without closing degree number entity, you cannot ensure that a full open and close has been done. That is why I check that the both entities are present. Else it could be useless. And moreover this has nothing to do with the issue itself.

I propose to take back the changes to fix the issue and all stuff around this. I will take your PR as is, it is a good starts.

bontiv:
I make the closing entity optional for TRV other than Sonoff TRVZB who don't have a closing entity, only a direct valve control like Eurotronics Comet on issue used by the user of issue #1660.

If a user use a Sonoff TRVZB and had not set a closing entity during the setup, I think he'll have also many problems for the TPI regulation...

Tested in my home, the main issue was resolved by the change on the lock. But... I'm agree with you. I don't understand why. I changed this because the only safe zones where initiate a lock are on init of a class or the main body of a file. Otherwise the state of the variable itself is unclear in parallel process (defined or not).

bontiv:
@jmcollin78 for the hvac mode, I use the lock to disable valve changes by the TPI algoritm. I think I forgot to change the hvac mode from off to heat at the begin of the calibration.

I will review it tomorrow. Sorry for that.

jmcollin78:
> I make the closing entity optional for TRV other than Sonoff TRVZB who don't have a closing entity, only a direct valve control like Eurotronics Comet on issue used by the user of issue #1660.

ok, I understand. The issue is that it will not work with Sonoff TRVZB. Then we have to explain that in documentation. Let's keep it.

> If a user use a Sonoff TRVZB and had not set a closing entity during the setup, I think he'll have also many problems for the TPI regulation...

No it works. The TRV choose its level between the min opening configured and its maximum (100%). It is not so precise as when  VTherm set opening and closing to constrain the valve position, but it is an interesting case (I think Maia does that to let the valve do its job).

> for the hvac mode, I use the lock to disable valve changes by the TPI algoritm. I think I forgot to change the hvac mode from off to heat at the begin of the calibration.

You will also have to change the `ThermostatClimate._async_climate_changed` to avoid changing the hvac_mode, hvac_action and setpoint. `_async_climate_changed` is used also for `ThermostatClimateValve`.

bontiv:
> You will also have to change the ThermostatClimate._async_climate_changed to avoid changing the hvac_mode, hvac_action and setpoint. _async_climate_changed is used also for ThermostatClimateValve.

Your're totally right!

I return in draft and fix the 3 remaining issues:
1. Change the forced state to "heat" instead of "off" (local change in my prod not commit)
2. Lock also climate change during calibration
3. Restore the valve position after the calibration

bontiv:
The last bug... I  don't understand why valve position is not restored after the calibration...

custom_components/versatile_thermostat/thermostat_climate_valve.py
```diff
@@ -482,10 +503,11 @@ async def _recalibrate_task():
 
             async with self._recalibrate_lock:
                 try:
-                    # Turn off vtherm
-                    _LOGGER.info("%s - Recalibration - Stopping VTherm and waiting for %s seconds", self, delay_seconds)
-                    await self.async_set_hvac_mode(VThermHvacMode_OFF, ignore_lock=True)
-                    await asyncio.sleep(delay_seconds)
+                    # Turn on vtherm if not already on
+                    if self.vtherm_hvac_mode != VThermHvacMode_HEAT:
```
jmcollin78, line 511: Sorry I was not clear. I think we need the Vtherm to be off but the underlying climate to be on. If VTherm is on, all mecanisms will work and will interfere with the recalibration.
So:
- restore the VTherm hvac_mode off,
- wait 5 sec (to wait for underlying climate to be off),
- turns on the underlying climate,
- wait delay_seconds
... continue

jmcollin78:
The overrides of the _async_xxx needs a unit test. Create a VTherm climate_over_valve, take the lock and try to call these methods.

jmcollin78:
Hello, because I need the update now to do a release, I will take the change directly. Thank you anymway for your work.

bontiv:
I'm really not satisfied with my work... I think it shouldn't have been merged.

One log write is missing in async_set_hvac_mode. I'm not writed the unit test for this feature and the last bug is still not resolved "Restore the valve position after the calibration".

With the inversion of the logic (VTherm off / under on), the bug have less impact but still present. If the user have set a closing degree under 100 and the VThem is off at the begin of the calibration, the valve position is not restored after calibration. The valve keep 0% opening / 100% closing of the last change by calibration and doesn't restore the `max_closing_degree`.

jmcollin78:
I have finished it yesterday. I wanted to put this fix in the release. 

> the last bug is still not resolved "Restore the valve position after the calibration".

You mean there is another bug yet ? I test it on my test Sonoff TRVZB and it has work fine. I have to make more test then with max_closing_degree set. Sorry I didn't understand this but only the miss of under on.

bontiv:
Wait and see. I don't think they are enough people who use max_closing_degree and make a calibration when there VTherm are off.

PR #1683, opened 2026-01-24
Refactor Auto-TPI cycle management to a passive model driven by thermostat control calls

KipK:
- Refactor Auto-TPI cycle management to a passive model driven by thermostat control calls.
- Removed internal timer-based loops in CycleManager and integrated learning/persistence into cycle completion hooks for better synchronization.
- Updated simulation tests to align with the new driving mechanism.

This is code I've taken back from SmartPI as timing management is far better and cleaner than previous one on auto_tpi.
New CycleManager class will also be used for SmartPI or any other algorithm that needs proper timing & cycle tracking.


KipK:
Je laisse en draft le temps de laisser tourner pour etre sur que rien n'est cassé.

KipK:
All good here.

PR #1686, opened 2026-01-24
Add startup completion tracking to prevent false triggers during initialization

madd0:
## Problem

During startup/reload, state is restored incrementally. Managers evaluating these intermediate values make decisions based on invalid data.

Observed in logs during a reload:

> target_temp=None → target_temp=7.0 → target_temp=17.0

With the intermediate `target_temp=7.0` (a default value) and `current_temp=17.3`, the auto-start/stop algorithm incorrectly determined heating wasn't needed—even though the actual target was 17.0°C.

This could cause:
- **Auto-start/stop**: Spurious stop/start events based on temporary default values
- **Safety manager**: False safety alerts from stale temperature timestamps  
- **Heating failure detection**: Incorrect failure detection from incomplete state

## Proposed Solution

Add an `is_startup_complete` flag to `BaseThermostat` that managers check before evaluating state:

1. Flag initialized to `False` in `__init__`
2. Flag set to `True` in `async_startup()` after state restoration completes, before the manager refresh loop
3. Managers skip evaluation if `False`, returning safe default values

Managers updated:
- `FeatureSafetyManager.refresh_state()`
- `FeatureHeatingFailureDetectionManager.refresh_state()`  
- `FeatureAutoStartStopManager.refresh_state()`

The startup check is placed after the `is_configured` check in each manager—no point logging "skipping during startup" if the feature isn't even enabled.

## Testing

Added unit tests verifying each manager skips evaluation when `is_startup_complete` is `False`.

### DIFF
--- a/tests/test_auto_start_stop_feature_manager.py
+++ b/tests/test_auto_start_stop_feature_manager.py
@@ -134,4 +134,5 @@ async def test_auto_start_stop_feature_manager_skips_during_startup(
 
     # refresh_state should not evaluate during startup and is_auto_stop_detected should remain False
     result = await auto_start_stop_manager.refresh_state()
+    assert result is False
     assert auto_start_stop_manager.is_auto_stopped is False


custom_components/versatile_thermostat/base_thermostat.py
```diff
@@ -552,6 +553,11 @@ async def async_startup(self, central_configuration):
                 self,
             )
 
+        # Mark startup as complete BEFORE manager refresh - state is now fully restored
+        # and managers can safely evaluate without seeing intermediate values
+        self._is_startup_complete = True
```
jmcollin78, line 558: You set _is_startup_complete here and then you initialize all manager. That mean all managers where initialized at this step.
And I know that all sensors are not ready even if this method is called.
So what you try to do is useless, not ?.

custom_components/versatile_thermostat/feature_auto_start_stop_manager.py
```diff
@@ -91,6 +91,9 @@ async def refresh_state(self) -> bool:
         if not self._is_configured or not self._is_auto_start_stop_enabled:
             _LOGGER.debug("%s - auto start/stop is disabled (or not configured)", self)
             self._is_auto_stop_detected = False
+        # Skip evaluation during startup to avoid acting on intermediate states
```
jmcollin78, line 94: refresh is not called before async_startup is done. So this is useless.

madd0:
I didn't spend _too much_ time on this, and the purpose of the PR is to have this discussion 🙂

I admit that Python is not a language that I use much, and don't have much experience with HA yet, so I may have misunderstood some the code, or the logs, but let me try to clarify what I'm seeing and the purpose of this fix.

You're correct that the explicit `manager.refresh_state()` loop happens after `async_startup()` sets up the state and that maybe not all sensors are ready at that point. However, the issue I observed wasn't with the explicit refresh loop, it was with **implicit callbacks** that fire during startup, probably when at least some of the sensors become ready.

When `async_added_to_hass()` runs, it registers state change listeners:

```python
self.async_on_remove(
    async_track_state_change_event(
        self.hass,
        [self._temp_sensor_entity_id],
        self._async_temperature_changed,
    )
)
```

During `async_startup()`, when the temperature sensor state is read and processed, it looks to me like this can trigger this callback, which can fire while the thermostat's state is not yet fully initialized.

In my logs, during an integration reload, I saw the auto-start/stop manager evaluating with `target_temp=None`, then `target_temp=7.0` (the min_temp fallback), before finally settling on the correct restored `target_temp=17.0`. The auto-start/stop algorithm was making decisions based on these intermediate states, which caused it to incorrectly trigger a stop.

Going through the code, I saw that the auto/start stop, safety and heating failure managers, can potentially be called in these conditions.

The idea of the `is_startup_complete` flag is to ensure that when `_async_temperature_changed` fires during the startup phase, the auto-start/stop manager (and other managers) skip their evaluation logic until the startup sequence explicitly marks itself complete, right before the intentional `refresh_state()` loop.

This is a defensive measure against the auto-start/stop algorithm acting on intermediate/uninitialized state during startup.

After running the code on my installation for a few days, I believe it has successfully prevented a few unintentional starts/stops. At the very least I see in the logs that "skipping auto start/stop evaluation during startup" shows, which means the state was being refreshed _before_ the flag was changed and its state completely restored.

That said, I did have at least one false start, so I'm also investigating whether there could be an issue because the `accumulated_error` is not restored from saved state, causing incorrect decisions after restart even with valid temperature values.

I'm not sure these changes are _totally_ useless, but you're the expert. Let me know if you have other questions or want to take a look at some of my logs. Or feel free to discard the PR if you're still not convinced.

jmcollin78:
Hello, I agree that auto-start/stop and heating anomaly detection can be fired before loading is complete. That is why each of the manager checks before doing something that all needed entities are valued and stop if it is not fulfilled.
So for me this is already taken into account in each manager. See the first lines of AutoStartStopManager.refresh_state:
```
@overrides
    async def refresh_state(self) -> bool:
        """Check the auto-start-stop and an eventual action
        Return True is auto stop is detected"""

        if not self._is_configured or not self._is_auto_start_stop_enabled:
            _LOGGER.debug("%s - auto start/stop is disabled (or not configured)", self)
            self._is_auto_stop_detected = False
```
and in `auto_start_stop_algorithm.py:
```
       if requested_hvac_mode is None or target_temp is None or current_temp is None:
            _LOGGER.debug(
                "%s - No all mandatory parameters are set. Disable auto-start/stop",
                self,
            )
            self._last_should_be_off = False
            return self._last_should_be_off
```

So for me all checks are already in the code. Maybe one is missing and a particular attention can be done  to those checks but basically principles are here and there is nothing more todo.

jmcollin78:
Hello, I rework all the code on this partocular subject. Follow the next release and let me know if it fixes your cases. I guess yes

PR #1687, opened 2026-01-24
corrected text formatting

silviopen:
During the translation of the text, the formatting of the JSON code was also incorrectly modified.

### DIFF
--- a/custom_components/versatile_thermostat/translations/it.json
+++ b/custom_components/versatile_thermostat/translations/it.json
@@ -1,791 +1,1011 @@
 {
-  "title": "Configurazione Versatile Thermostat",
-  "config": {
-    "flow_title": "Configurazione Versatile Thermostat",
-    "step": {
-      "user": {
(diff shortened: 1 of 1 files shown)


silviopen:
Hello,
To avoid errors in the future, I am looking for a tool to update .json files, but I haven't found anything simple and inexpensive.
I used poedit, but (in the free version) it doesn't allow me to compare two files and add missing entries.
Can you suggest something useful?

jmcollin78:
Yes, to update json, interact with Github, you should use VSCode.

silviopen:
I am learning to use VSCODE to add missing entries between languages, and I must say that it is very convenient.
I am not sure which extension can help me speed up translation without having to write terms that have already been used every time, as I do with POedit.

jmcollin78:
The extension that will help you for translation is definitively Github Copilot (an AI). If you can have Claude or Chatgpt or Gemini their are really helpful for translations and uniformisation.

PR #1702, opened 2026-01-28
Fix README missing Heating Failure Detection feature

Reifircax:


jmcollin78:
It seems there is an validation issue.

jmcollin78:
ok it seems it is not your change. I will have a look.

PR #1743, opened 2026-02-08
Support temperature attribute and weather sensors

mariusangelmann:
Make external temperature handling more robust and allow weather entities as external sensors.

- auto_tpi_manager.py: import ATTR_TEMPERATURE and prefer reading temperature from either the entity state or the temperature attribute when computing outdoor temperature (skip unknown/unavailable and non-numeric values).
- base_thermostat.py: introduce _extract_temperature_value to centrally extract a finite temperature from state or attributes (optionally allowing the temperature attribute). Use this helper in _async_update_temp and _async_update_ext_temp (ext update now returns bool). Add ext_current_temperature_entity_id to the restored state. Improve logging when external sensor does not provide a usable temperature.
- config_schema.py: include the weather domain in the external temperature entity selector so weather entities can be chosen as external temp sensors.

These changes improve resilience to sensors that expose temperature in attributes and expand supported entity types for external temperature.

### DIFF
--- a/custom_components/versatile_thermostat/auto_tpi_manager.py
+++ b/custom_components/versatile_thermostat/auto_tpi_manager.py
@@ -8,7 +8,7 @@ import statistics
 from datetime import datetime, timedelta
 from typing import Optional
 from homeassistant.util.unit_conversion import TemperatureConverter
-from homeassistant.const import UnitOfTemperature
+from homeassistant.const import UnitOfTemperature, ATTR_TEMPERATURE
 from dataclasses import dataclass, asdict, field
 
 import asyncio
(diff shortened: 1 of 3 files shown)


jmcollin78 (commented):
Seems good for me
But you have a conflict now (sorry I push a fix just before).
Please add a simple unit test that creates a VTherm with a weather entity. See example in test_start.py.

jmcollin78:
Hello @mariusangelmann,

The more you wait to resolve the conflict, the more your PR will have conflicts. You are missing the R9.0.0 release.
And you should add a test unit to proove that your change works.

jmcollin78:
Stale PR.... So sad.

Issue #1679, opened 2026-01-23
Rooms not reaching temperature since 8.6.1 - regression on self-regulation ?


jmcollin78:
### Discussed in https://github.com/jmcollin78/versatile_thermostat/discussions/1643

<div type='discussions-op-text'>

<sup>Originally posted by **jkbkster** January 18, 2026</sup>
None of my 9 rooms are reaching their temperature after 8.6.0 update. 
Nothing apart from the update has been changed. 

I just need the valves to open fully like before - now they only reach 20-30% . 

What do I do ?</div>

bontiv:
This issue is older than the 8.6.1. I'm not sure it's a regretion. I will check it.

jmcollin78:
If I look at the very end of the post, it seems really related to 8.6.1. I ask for a rollback to be sure.

jmcollin78:
@jkbkster please do the rollback and let us know if it is related to 8.6.1. This is important for us to know. Thank you

jkbkster:
> @jkbkster please do the rollback and let us know if it is related to 8.6.1. This is important for us to know. Thank you

I've just downgraded to 8.5.1 and immediately radiators went to 90-100% on heating demand.

My feeling is the issues were introduced in 8.6, and escalated in 8.6.1

Will have more data tmrw

jmcollin78:
ok thank you for this test.
You are in `over_climate` without valve regulation is that right ?

I will check all changes between 8.5.1 and 8.6.x. to try to find something related. Keep the 8.5.1 then please.

jkbkster:
Yes; this is config from one of the rooms:

```
hvac_modes:
  - "off"
  - heat
min_temp: 17
max_temp: 30
target_temp_step: 0.1
preset_modes:
  - none
  - frost
  - eco
  - comfort
  - boost
current_temperature: 20.2
temperature: 20.5
hvac_action: heating
preset_mode: none
hvac_mode: heat
ema_temp: 20.12
specific_states:
  is_on: true
  last_central_mode: Auto
  last_update_datetime: "2026-01-24T09:37:50.522038+01:00"
  ext_current_temperature: 0
  last_temperature_datetime: "2026-01-24T08:57:46.200072+01:00"
  last_ext_temperature_datetime: "2026-01-24T09:36:10.488967+01:00"
  is_device_active: true
  device_actives:
    - climate.danfoss_etrv0100_termostat
  nb_device_actives: 1
  ema_temp: 20.12
[...]

PR #1751, opened 2026-02-10
Change the dtemp behaviour

bontiv:
Before:

dtemp = target_temp - self._regulated_target_temp

After:

dtemp = target_temp - (under.last_sent_temperature if under.last_sent_temperature else 0)

Change impact
This PR changes the calculation behavior of the dtemp (temperature difference) within the self-regulation framework:

Old logic: The difference was calculated relative to the stored regulated target temperature (self._regulated_target_temp), which did not reflect the last temperature sent to the underlying device.

New logic: The difference is now calculated relative to the last temperature actually sent to the underlying device (under.last_sent_temperature).

Bug discovered during the log exploitation of #1679

### DIFF
--- a/custom_components/versatile_thermostat/underlyings.py
+++ b/custom_components/versatile_thermostat/underlyings.py
@@ -968,11 +968,15 @@ class UnderlyingClimate(UnderlyingEntity):
         if ClimateEntityFeature.TARGET_TEMPERATURE in self.supported_features:
             data["temperature"] = target_temp
 
-        await self.hass_services_async_call(
-            CLIMATE_DOMAIN,
-            SERVICE_SET_TEMPERATURE,
-            data,
-        )
+        try:
+            await self.hass_services_async_call(
+                CLIMATE_DOMAIN,
+                SERVICE_SET_TEMPERATURE,
+                data,
+            )
+        except Exception as ex:
+            _LOGGER.error("%s - Error while sending set_temperature: %s", self, ex)
+            raise ex
 
         self._last_sent_temperature = target_temp
         _LOGGER.debug("%s - Last_sent_temperature is now: %s", self, self._last_sent_temperature)


jmcollin78 (approved):
Did you experiment such error ?

Panoramiac:
Is this the reason why it is currently overshooting when using Ausgleich der Innentemperatur als Basis? Happy that the provided logs already helped to fix this issue ;-).

jmcollin78:
Visible in preview : https://github.com/jmcollin78/versatile_thermostat/releases/tag/9.0.0.beta1

Feedback appreciated before the official release this week-end. Thank you

bontiv:
Hello @Panoramiac,

Yes, please test the beta. The explanation:

With the old dtemp:

```
dtemp = target_temp - regulated_target_temp
target_temp = new_regulated_temp + offset_temp
offset_temp = device_temp - current_temperature

target_temp = current_temperature + pi_algo_offset + device_temp - current_temperature
target_temp = new_regulated_temp + device_temp - current_temperature

dtemp = new_regulated_temp + device_temp - current_temperature - regulated_target_temp
dtemp = (new_regulated_temp - regulated_target_temp) + (device_temp - current_temperature)
```

So the test to skip the underlying device setpoint change:

```
if abs(dtemp) < auto_regulation_dtemp:
   skip
```

Become:
```
(new_regulated_temp - regulated_target_temp) + (device_temp - current_temperature) < auto_regulation_dtemp
```

### Step 1, heating.

Target at 21, underlying is set to 26 (+2 from `(device_temp - current_temperature)` and +3 from PI Algorithm).

### Step 2, cooing

Reduce the target to 19.

If the PI_algorithm need -2C and there are +2C between the temperature from the underlying climate sensor and the room sensor, the new target is not send to the underlying device. So the room keep on the state 1 heating.

---

The new test compare the last target temperature sent to the underlying device to the target_temp which is the wanted target (sended to underlying at line 256):

```python
await under.set_temperature(
    target_temp,
    self._attr_max_temp,
    self._attr_min_temp,
)
```

So we only skip if the difference between the last setpoint and wanted setpoint is under the auto_regulation_dtemp.

PR #1756, opened 2026-02-10
remove useless double calculation of on_time_sec/off_time_sec in thermostat_switch

KipK:
from yesterday night discussion.

Use real previously calculated on/off_time_sec sent to base_thermostat
set off_time_sec to cycle time when OFF for sensor logic.


jmcollin78 (approved):

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/releases/tag/9.0.0.beta1

PR #1764, opened 2026-02-13
fix(underlyings): fix race condition between keep_alive and start_cycle

KipK:
Fixes incorrect duty cycles when keep_alive is enabled. The _should_be_on flag was temporarily set to False at cycle start, creating a race window where keep_alive callback could incorrectly turn off the device before the scheduled turn_on executed.

Now sets _should_be_on to the correct intended state in each branch before any async operations, eliminating the race condition.


KipK:
c'était le bug de gael d'hier. 
Si keep_alive set, a 5mn, pour un cycle de 10mn, a 100% power ca coupait toutes les 5mn

PR #1773, opened 2026-02-16
Avoid sending offset updates if new value is equal to current one

mircoboschi:
Hello,
i'm using versatile_thermostat with on/off TRVs.
I see with zigbee2mqtt that versatile_thermostat sends temperature offset updates even if offset is already fine as-is (confirmed reading versatile_thermostat debug logs).
This simple patch avoid pushing new offset to underlying if not needed.


jmcollin78 (approved):
Thank you it is helpful indeed

PR #1825, opened 2026-02-28
Change logo

jmcollin78:
On an original idea of @Krzysztonek


jmcollin78:
Thank you @Krzysztonek for the original idea. Finally I will keep this more flat and modern version.

PR #1827, opened 2026-03-01
feat: add stuck TRV valve detection and auto-recovery

ToRvaLDz:
## Summary

When using `over_climate` thermostats backed by TRV valves (e.g., Zigbee radiator valves), a common real-world issue is the valve **physically getting stuck in the "heating" position** even after the room temperature has exceeded the target. This results in wasted energy, uncomfortable overheating, and no built-in way for the user to know it's happening until they notice the room is too warm.

This PR adds a new optional feature — **Stuck Valve Detection** — that:

1. **Detects** when an underlying TRV reports `hvac_action: heating` while `current_temperature > target_temperature` for a configurable delay (default: 120 seconds)
2. **Automatically cycles** the valve off and back on to attempt to physically unstick it
3. **Limits retries** to a configurable maximum number of cycles (default: 3), after which it fires a Home Assistant event and logs a warning so the user can take manual action
4. **Fires events** (`versatile_thermostat_stuck_valve_event`) for easy integration with HA automations and notifications

### Why this is needed

The existing **Heating Failure Detection** feature covers the opposite scenario (thermostat is heating but temperature is not rising). There is currently **no mechanism** to detect or recover from a valve that remains open when it shouldn't be. Users with multiple TRV-based zones have reported this as a recurring problem, especially with cheaper Zigbee valves that occasionally lose sync with their internal motor position.

## Changes

- **New file:** `feature_stuck_valve_detection_manager.py` — implements the detection logic as a `BaseFeatureManager`, following the same pattern as existing feature managers (heating failure, motion, window, etc.)
- **`const.py`** — new configuration constants and `STUCK_VALVE_EVENT` event type
- **`config_schema.py`** — schema definitions for the new configuration options
- **`config_flow.py`** — new UI step `stuck_valve_detection` with configurable delay and max cycles
- **`thermostat_climate.py`** — registers the manager in `ThermostatOverClimate`
- **`base_thermostat.py`** — calls `refresh_state()` on the stuck valve manager during each update cycle
- **`strings.json`** + **all 9 translation files** (en, it, fr, de, cs, el, pl, ru, sk) — full UI translations for the new feature

## Configuration options

| Option | Default | Description |
|--------|---------|-------------|
| `use_stuck_valve_detection_feature` | `false` | Enable/disable the feature |
| `stuck_valve_detection_delay_sec` | `120` | Seconds to wait before considering a valve stuck |
| `stuck_valve_max_cycles` | `3` | Max off/on cycles before giving up and firing an alert event |

## Test plan

- [ ] Enable the feature on an `over_climate` VTherm with a TRV
- [ ] Simulate a stuck valve (manually set TRV to heating while room temp > target)
- [ ] Verify detection triggers after the configured delay
- [ ] Verify the off/on cycle is performed and logged
- [ ] Verify the event `versatile_thermostat_stuck_valve_event` is fired
- [ ] Verify max cycles limit is respected
- [ ] Verify feature does nothing when disabled
- [ ] Verify feature is ignored on non-`over_climate` thermostats

### DIFF
--- a/custom_components/versatile_thermostat/feature_heating_failure_detection_manager.py
+++ b/custom_components/versatile_thermostat/feature_heating_failure_detection_manager.py
@@ -334,7 +334,7 @@ class FeatureHeatingFailureDetectionManager(BaseFeatureManager):
         """Diagnose the root cause of a failure by checking valve underlyings.
 
         For thermostats with valve underlyings (over_valve or over_climate_valve),
-        compares the commanded state (should_device_be_active) with the real state
+        compares the requested state (should_device_be_active) with the real state
         (is_device_active) to determine if a stuck valve is the cause.
 
         Args:
@@ -369,14 +369,14 @@ class FeatureHeatingFailureDetectionManager(BaseFeatureManager):
                 stuck_valves.append({
                     "entity_id": under.entity_id,
                     "type": "valve_stuck_closed",
-                    "commanded": "open",
+                    "requested": "open",
                     "actual": "closed",
                 })
             elif not should_active and is_active:
                 stuck_valves.append({
                     "entity_id": under.entity_id,
                     "type": "valve_stuck_open",
-                    "commanded": "closed",
+                    "requested": "closed",
                     "actual": "open",
                 })
 
--- a/tests/test_heating_failure_detection.py
+++ b/tests/test_heating_failure_detection.py
@@ -1026,7 +1026,7 @@ async def test_diagnose_root_cause_valve_stuck_closed(hass: HomeAssistant):
     assert result["root_cause_entity_id"] == "number.trv_living_room"
     assert len(result["root_cause_details"]) == 1
     assert result["root_cause_details"][0]["type"] == "valve_stuck_closed"
-    assert result["root_cause_details"][0]["commanded"] == "open"
+    assert result["root_cause_details"][0]["requested"] == "open"
     assert result["root_cause_details"][0]["actual"] == "closed"
 
 
@@ -1052,7 +1052,7 @@ async def test_diagnose_root_cause_valve_stuck_open(hass: HomeAssistant):
     assert result["root_cause_entity_id"] == "number.trv_bedroom"
     assert len(result["root_cause_details"]) == 1
     assert result["root_cause_details"][0]["type"] == "valve_stuck_open"
-    assert result["root_cause_details"][0]["commanded"] == "closed"
+    assert result["root_cause_details"][0]["requested"] == "closed"
     assert result["root_cause_details"][0]["actual"] == "open"
 
 


jmcollin78:
Hello @ToRvaLDz ,

Very interesting PR thank you. But ... this case is already detected by the "Heating Failure Detection". It detects heating failure but also cooling failure (the temperature drops but it shouldn't).
I guess this part of your PR is redondant.

What is new is the try to solve the issue by itself. For using the "Heating Failure Detection", I can ensure you there false positive detection. Suppose you have a wood stove, then all the TRV will alerts with a heating detection failure.
I pretty sure it is not suitable that they all do a cycle of recovery because they have no defect. It will just drawn the battery.

You should ask me before doing such implementation. I know, you don't do it by hand but it is IA generated, but it is still time and resources consumption if finally, it is not kept.

Nevertheless, I appreciated the effort and the conformance to the VTherm architecture.

What do yo think ?

ToRvaLDz:
Hi @jmcollin78,

Thank you for the detailed feedback — you raise very valid points.

**You're absolutely right about auto-cycling recovery**: it would cause false positives with wood stoves and drain TRV batteries unnecessarily. I agree this should be removed.

However, I believe the two features detect **fundamentally different problems**:

- **Heating Failure Detection** monitors **temperature vs. on_percent** — a thermal anomaly (e.g., heating is on but temperature doesn't rise)
- **Stuck Valve Detection** would monitor **valve position vs. commanded state** — a mechanical/hardware failure (e.g., VTherm commands the valve to close, but `current_valve_opening` still reports it open)

The key insight is: **wood stoves don't have TRV valves with position indicators**. By using `current_valve_opening` instead of temperature, false positives from external heat sources are completely eliminated — only entities with actual valve position reporting would be checked.

I also noticed that VTherm already has the building blocks for this:
- `should_device_be_active` vs `is_device_active` in `UnderlyingValve`
- `check_initial_state()` already detects this exact mismatch — but **only once at boot**, not continuously

The idea would be to extend this check continuously: if `should_device_be_active != is_device_active` persists for a configurable delay, fire a notification event. The user could then decide what to do — either handle it via an automation or optionally let VTherm attempt to re-send the command to the valve.

**Two possible implementation paths:**

1. **Extend Heating Failure Detection** — add valve position mismatch as a third check alongside heating/cooling failure, within the existing manager
2. **Keep it as a separate feature** — this would allow users to enable/disable it independently from temperature-based failure detection

Which approach would you prefer? I'm happy to rework the implementation either way.

jmcollin78:
A stuck valve is "just" one of the root of a Heating detection failure. So for me, it is the same feature but you are exploring the root cause analysis. So I see your feature request as an extension of the "Heating detection failure".

Did you notice that the "Heating detection failure" detects "heating failure" (the valve is stuck open) as well as the "cooling failure" (the valve is stuck close). See [here](https://github.com/jmcollin78/versatile_thermostat/blob/main/documentation/en/feature-heating-failure-detection.md)

So what is pertinent is that for a thermostat with valve (`over_valve` or `over_climate_valve`) if a heating/cooling detection failure is detected, you can try to diagnose a valve stuck comparing the expected valve position and the real valve position. Then you PR becomes a root cause analysis of a failure detection and that is particulary pertinent. I suggest to add a root-cause field in the failure message so that it can be catch by an automation or whatever.

Be careful that is_device_active works on the real valve position and should_be_active is working on the brut valve position (ie the result of calculation). Between both there is calculation (min/max/open/close/threshold). So i'm not sure it is totally pertinent to compare the both. But I have to admit, it has to be verified.

I keep thinking it is not a good idea to try to fix the issue automaticaly. It is better to notify and let each user do what they want to do. If they want to implement an full on/off in the valve, fine, but it is its responsability.

For instance, for forcing a full on/off without disturbing the Vtherm, you have the recalibration action which does exactly this. I suggest you add this as an example in the documentation of the feature.

ToRvaLDz:
Hi @jmcollin78,

Thank you for the very insightful guidance. I fully agree with the approach: this should be a **root cause analysis extension** of the existing Heating Failure Detection, not a separate feature.

**Regarding `should_device_be_active` vs `is_device_active`:**

You're right that there are transformations between the brut and real values (`OpeningClosingDegreeCalculation` with min/max/opening/closing degrees). However, for stuck valve detection I believe the **boolean comparison** is still valid:

- VTherm commands 0% → `should_device_be_active = False` (regardless of threshold)
- Valve physically stuck at 50% → `is_device_active = True` (regardless of threshold)
- Mismatch detected → root cause: `valve_stuck_open`

The thresholds differ between the two (`_opening_threshold` vs `100 - _max_closing_degree` for `UnderlyingValveRegulation`), but since we're comparing intent ("should be OFF") vs reality ("is ON"), the boolean level should be reliable. That said, if you'd prefer comparing the transformed percentage values directly instead of the boolean properties, I'm happy to adjust.

**Proposed implementation:**

When a heating or cooling failure is detected on a thermostat with valve underlyings (`is_over_valve` or `_underlyings_valve_regulation`), add root cause analysis by checking each underlying's `should_device_be_active` vs `is_device_active`. Then extend the existing event payload with:

```python
{
    # ... existing fields ...
    "root_cause": "valve_stuck_open",  # or "valve_stuck_closed" or "unknown"
    "root_cause_entity_id": "climate.trv_living_room",
}
```

This way automations can react specifically to stuck valves while the existing temperature-based detection remains untouched.

**No automatic fix** — I agree it's better to notify and let the user decide. I noticed the `recalibrate_valves` service already does exactly what a user would want in this scenario (full open/close cycle). I'll add a note in the documentation suggesting users can call `recalibrate_valves` from an automation triggered by the failure event with `root_cause: valve_stuck_*`.

Should I go ahead and rework the PR with this approach?

jmcollin78:
> Should I go ahead and rework the PR with this approach?

yes please, you can proceed.
One more point, if the heating failure is not due to a valve stuck then add a default cause like "not identified". Maybe I will be able to add more diagnostic root cause in the future.

Of course to be acceptable, I need some unit tests showing the cases when it detects the valve stuck and when not. There is already tests you can rely on.

jmcollin78:
Hello @ToRvaLDz, do you intend to finish this PR ? If not no problem, I will just close it.
The more you wait and the more the rebase will be difficult.

ToRvaLDz:
Hi @jmcollin78,

Sorry for the delay — I've been busy with other projects. Here's the reworked PR as discussed.

**Removed**: the entire separate "Stuck Valve Detection" feature (separate manager, config flow step, translations, auto-cycling logic).

**Added**: root cause analysis as an extension of the existing Heating Failure Detection. When a heating or cooling failure is detected on a thermostat with valve underlyings, the manager now compares `should_device_be_active` vs `is_device_active` on each underlying valve to diagnose if the failure is caused by a stuck valve.

The failure event now includes three new fields:

```python
{
    # ... existing fields unchanged ...
    "root_cause": "valve_stuck_closed",       # or "valve_stuck_open" or "not_identified"
    "root_cause_entity_id": "number.trv_sala",
    "root_cause_details": [{"entity_id": "...", "type": "...", "commanded": "...", "actual": "..."}]
}
```

The same information is also exposed in the entity's custom attributes when a failure is active.

**What it does NOT do**: no automatic fix, no auto-cycling. Users can set up an automation on the event and optionally call `recalibrate_valves` if they want to attempt recovery.

**Tests**: 12 new unit tests covering all scenarios (stuck closed, stuck open, no valve underlyings, multiple valves, unavailable state, attributes).

All 34 tests pass (22 existing + 12 new).

jmcollin78:
Is that still in draft ?

custom_components/versatile_thermostat/feature_heating_failure_detection_manager.py
```diff
@@ -325,12 +330,90 @@ def _reset_tracking(self):
...
+            is_active = under.is_device_active
+
+            if should_active is None or is_active is None:
+                continue
+
+            if should_active and not is_active:
+                stuck_valves.append({
+                    "entity_id": under.entity_id,
+                    "type": "valve_stuck_closed",
+                    "commanded": "open",
```
jmcollin78: requested instead of commanded not ? (I'm not sure but I use the requested word may times)
Replace everywhere.

ToRvaLDz: Hi @jmcollin78, good catch, I've renamed commanded to requested everywhere (code, docstrings, and tests). Also marked the PR as ready for review. Let me know if anything else needs adjusting!

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/releases/tag/9.3.0

PR #1841, opened 2026-03-06
IT update to align it with the FR version

silviopen:
IT update to align it with the FR version


jmcollin78:
Hello, validation fails on your PR. Please fix and I would be pleased to merge it.

PR #1855, opened 2026-03-13
fix(cycle_scheduler): close race window with _is_starting guard, add cycleScheduler.shutdown() to properly cancel timers

KipK:
- fix(cycle_scheduler): close race window with _is_starting guard:
 Add _is_starting flag to keep is_cycle_running=True during the window between cancel_cycle() returning and _start_cycle_switch() installing new timers.

Without this guard, a concurrent start_cycle(force=False) could see is_cycle_running=False, create duplicate timers, and leak them — causing cascading premature cycle ends and double energy counting until HA restart.

- fix(cycle_scheduler): cancel timers on entity removal to prevent stale callbacks
 Add CycleScheduler.shutdown() to cancel pending timers synchronously, and call it from BaseThermostat.remove_thermostat() during entity removal.

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -438,6 +438,11 @@ class BaseThermostat(ClimateEntity, RestoreEntity, Generic[T]):
 
         self.stop_recalculate_later()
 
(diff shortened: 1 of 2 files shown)


jmcollin78 (approved):
Seems legit

PR #1863, opened 2026-03-16
Fix anchor name

cnotin:
Here's the correct URL that I noticed: https://www.versatile-thermostat.org/en/docs/troubleshooting/#why-is-my-versatile-thermostat-going-into-safety-mode


jmcollin78 (approved):
Thank you for the update !

PR #1865, opened 2026-03-16
Update Italian translations in it.json

silviopen:


silviopen:
After adding the missing lines to the Italian version, I checked the file with POEdit, and it flagged some sentences as missing a period.
I corrected only a few of them because I noticed that even in the French version, some sentences ended with a period and others did not, so I began to wonder if there was a specific rule.
If you can confirm that the period should be placed exactly as in the French version, I have no problem correcting this issue as well.

Il 17 marzo 2026 08:23:05 CET, Jean-Marc Collin ***@***.***> ha scritto:

silviopen:
Hi, 
I'm checking the differences between the FR and IT files and noticed that there's a typo in the French version on lines 202 and 583.
“motion_off_delai” should be “motion_off_delay”.
I usually refer to the fr.json file instead of the en.json file because it seems to be the most up-to-date, but if I’m wrong, please let me know.

jmcollin78:
> I usually refer to the fr.json file instead of the en.json file because it seems to be the most up-to-date, but if I’m wrong, please let me know.

You are right. They should be at the same level, but I do french first.

Thank you for the report.

Issue #1867, opened 2026-03-17
Bug: hvac_action simulation using wrong data/no heat action when needed (continuation of #1852)


gege32:
## Version of the custom_component

>9.0.0 (8.6.0 didn't have this issue yet)

## Configuration

My VTherm attributes are the following (before changing target temp):
```yaml
hvac_modes: off, heat
min_temp: 7
max_temp: 30
target_temp_step: 0.5
preset_modes: none, frost, eco, comfort, boost
current_temperature: 19.2
temperature: 24
hvac_action: idle
preset_mode: none
hvac_mode: heat
ema_temp: 19.27
is_ready: true
specific_states: 
is_initialized: true
is_on: true
last_central_mode: Heat only
last_update_datetime: '2026-03-12T20:47:06.595316+01:00'
ext_current_temperature: 9.6
last_temperature_datetime: '2026-03-12T20:45:27.284835+01:00'
last_ext_temperature_datetime: '2026-03-12T20:47:06.583462+01:00'
should_device_be_active: false
is_device_active: false
device_actives: []
nb_device_actives: 0
ema_temp: 19.27
temperature_slope: -0.6
hvac_off_reason: null
total_energy: 0
last_change_time_from_vtherm: '2026-03-12T20:45:30.809821+01:00'
is_sleeping: false
is_locked: false
is_recalculate_scheduled: true
not_initialized_entities: []
messages: []

configuration: 
ac_mode: false
type: over_climate
is_controlled_by_central_mode: true
target_temperature_step: 0.5
timezone: Europe/Amsterdam
temperature_unit: °C
is_used_by_central_boiler: true
max_on_percent: null
have_valve_regulation: false
cycle_min: 30

preset_temperatures: 
frost_temp: 10
eco_temp: 17
boost_temp: 23
comfort_temp: 20
frost_away_temp: 10
eco_away_temp: 16
boost_away_temp: 21
comfort_away_temp: 18

current_state: 
hvac_mode: heat
target_temperature: 24
preset: none

requested_state: 
hvac_mode: heat
target_temperature: 24
preset: none

is_presence_configured: true
presence_manager: 
presence_sensor_entity_id: input_boolean.somebody_home_sensor
presence_state: 'on'

is_power_configured: false
power_manager: 
device_power: 1
mean_cycle_power: 0

is_motion_configured: false
is_window_configured: true
is_window_auto_configured: false
window_manager: 
window_state: 'off'
window_auto_state: unavailable
[...]

PR #1868, opened 2026-03-17
fix: hvac_action simulation has missing underlying_target_temperature #1867

gege32:
# Fix hvac_action simulation (#1867)

In underlyings.py underlying_target_temperature property reads "target_temperature" attribute. This attribute is (sometimes?) called "temperature". The code now also checks "temperature" attribute when the other is not available. More details about this in the related issue

```python
    @property
    def underlying_target_temperature(self) -> float:
        """Get the target_temperature"""
        value = self.get_underlying_attribute("target_temperature")
        if value is not None:
            return value
        return self.get_underlying_attribute("temperature")
```


jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/releases/tag/9.3.0

PR #1892, opened 2026-03-23
Updating IT translation

silviopen:


bontiv:
Hi @silviopen,

You translated the wrong file. You changed the en.json instead of it.json.

silviopen:
I apologize.
It's the right file with the wrong name.
I renamed the file to run some tests and forgot to change the name back.
I'll fix the error tonight.
Thanks for letting me know.

Issue #1541, opened 2026-01-05
[Feature request] Allow manual switch override


jmcaracoche:
Hello, first of all thank you to the community for building  such awesome and useful component.

I faced an issue with my use case so I plan to contribute on this feature (I already have the solution on testing and I will submit the changes once everything is polished)

My Use case:
I have an IR blaster and smart switch with power consumption to control a split AC unit. So I built a switch combining both (IR sensor and power meter)

Therm works OK but in some context during the off-cycle... my child got hot and then they turned on the AC manually via the remote control and couple of seconds after VTherm tuns is off due to the safety mechanism that was built if to avoid under to over cooling.

The problem with is I got resistance in the family on implementing this automations at home and they want to get the things manual if they want. So as I want to have as much things automated as I want, I decided to jump into the code to implement the following use cases.

1. VTherm controlling AC (ON), AC turned OFF externally: VTherm detects the switch turning OFF while it thinks it should be ON, so it turns VTherm OFF.

2. VTherm OFF, AC turned ON externally: VTherm stays OFF and allows the switch to remain ON if allow_manual_override is enabled.

3. VTherm ON (Idle), AC turned ON externally:  VTherm detects the switch turning ON while it was not requesting it, so it turns VTherm OFF to yield to manual control.

4. AC ON manually, then VTherm turned ON: When you switch VTherm to COOL/HEAT, it starts its control algorithm and takes over, either keeping the switch ON (if temp requires it) or turning it OFF (if temp is reached).

This is my first time contributing so I plan to create a branch for this and then the PR

jmcollin78:
Hello,

Please before doing PR, I propose to discuss about it. You will loose your time if your PR will break things. VTherm is now a complex integration. PR often breaks things because there is principles to follow that IA don't have in mind (they have no mind).

And I spend more and more time reading PR done by IA which are finally not applicable. My time is precious.

So let discuss about before publishing any PR, please.

I have first to try to understand what problem you try to solve... Ultra specific cases will not be accepted.

jmcaracoche:
sure I'l do that. I will explain the changes I made which are very straightforward.

jmcollin78:
Hello, I read carefully. Here are my questions / comments:

> VTherm works OK but in some context during the off-cycle... my child got hot and then they turned on the AC manually via the remote control and couple of seconds after VTherm tuns is off due to the safety mechanism that was built if to avoid under to over cooling.

[...]

PR #1901, opened 2026-03-26
feat: allow manual overrides on switch thermostat (#1541)

jmcaracoche:
## Summary

- Adds a new per-switch-thermostat option **`allow_manual_override`** (default `False`) that lets the physical switch or AC remote take precedence over VTherm's automatic control without VTherm fighting back.

## Use cases covered

| # | Scenario | Expected behaviour |
|---|----------|--------------------|
| UC1 | VTherm **ON**, switch turned **OFF** manually while in ON cycle | VTherm turns itself **OFF** (does not restart the switch) |
| UC2 | After UC1, VTherm is OFF | VTherm stays OFF and never turns the switch back on |
| UC3 | VTherm **OFF**, switch turned **ON** manually | VTherm stays OFF and does not interfere |
| UC4 | VTherm **ON but idle** (off-part of cycle), switch turned **ON** manually | VTherm turns itself OFF; skip-turn-off guard prevents it from immediately turning the switch back off |
| UC_WIN | Window **open** while device is active (manual override) — or switch turned ON while window is open | Configured window action is **always enforced**, overriding manual state |

## Implementation details

- **`CONF_ALLOW_MANUAL_OVERRIDE`** — new boolean config flag in `const.py` and `config_schema.py`.
- **`ThermostatOverSwitch._async_switch_changed`** — detects manual state changes and triggers the appropriate response. Window-open check takes unconditional priority.
- **`UnderlyingSwitch.set_hvac_mode`** — honours `_skip_turn_off_on_next_hvac_off` flag so the switch is not turned off during a manual-takeover transition.
- **`TPIHandler`** — respects `_allow_manual_override` at end-of-cycle; window detection still forces the device off regardless.
- **`FeatureWindowManager.update_window_state` / `_apply_window_action_on_manual_override`** — applies the window action when VTherm is already OFF but the underlying device is active, and flushes the updated `window_state` to HA's state machine.
- **Translations** (en, fr, de, cs, el, it, pl, ru, sk) and **reference documentation** updated.
- **Full test coverage** in `tests/test_manual_override_switch.py` (UC1–UC4, UC_WIN, feature-disabled path).

## Test plan

- [ ] All 624 existing tests pass (`pytest tests/ -q`)
- [ ] UC1: VTherm ON + remote turns switch OFF → VTherm goes OFF
- [ ] UC2: VTherm stays OFF after UC1, switch is not turned back on
- [ ] UC3: VTherm OFF + remote turns switch ON → VTherm does not interfere
- [ ] UC4: VTherm idle + remote turns switch ON → VTherm goes OFF, switch stays ON
- [ ] UC_WIN: window opens while device is manually active → window action fires
- [ ] UC_WIN reverse: switch turned ON while window is already open → window action fires immediately
- [ ] `allow_manual_override = False` (default) → behaviour is identical to before this PR


bontiv:
Note: See the discussion #1544

PR #1907, opened 2026-03-28
Fix formatting due to unecessary markdown block

cnotin:
Removed unnecessary markdown code block so it renders correctly, instead of showing the markdown

Here's how it looks like currently:
* on the website:
 
  image

* on github event with Preview enabled:

  image

### DIFF
--- a/documentation/cs/feature-advanced.md
+++ b/documentation/cs/feature-advanced.md
@@ -1,4 +1,3 @@
-````markdown
 # Pokročilá konfigurace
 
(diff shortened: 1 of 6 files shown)


cnotin:
I've fixed it in translations too, hoping that it won't disturb any automated process you may have. I let you know and say otherwise!

cnotin:
CI is red in the unit tests but it seems weird because I didn't change the code. Perhaps a flaky test? But I can't re-run it to verify

Issue #1722, opened 2026-02-02
Missing State Restoration in TimedPresetManager


jmcollin78:
> 

### Discussed in https://github.com/jmcollin78/versatile_thermostat/discussions/1720

<div type='discussions-op-text'>

<sup>Originally posted by **TTLucian** February  2, 2026</sup>
TL;DR
Timed presets (like Boost) currently remain active indefinitely after a restart because the manager does not implement a restoration hook. Adding a restore_state method allows the manager to re-populate its internal timer from existing entity attributes and resume the countdown.

While BaseThermostat.get_my_previous_state (line 475) correctly iterates through all feature managers to call manager.restore_state(old_state), this method is currently missing from FeatureTimedPresetManager.py.

Consequently, the _timed_preset_end_time and the async_track_point_in_time task are lost during a Home Assistant restart. Since the timer is never rescheduled, the thermostat remains "stuck" in the temporary preset.

By implementing the restore_state method in FeatureTimedPresetManager.py, we can re-populate the manager's state using the data already persisted in the entity attributes.

```Python
@callback
    def restore_state(self, last_state: Any):
        """Implement the restoration hook to re-populate timed presets after restart."""
        
        # 1. Retrieve the persistence dictionary from attributes
        # Matches the key used in add_custom_attributes
        manager_attr = last_state.attributes.get("timed_preset_manager")
        if not manager_attr or not manager_attr.get("is_active"):
            return

        end_time_str = manager_attr.get("end_time")
        preset_str = manager_attr.get("preset")

        if not end_time_str or not preset_str:
            return

        try:
            # 2. Re-parse the end_time and preset mode
            end_time = datetime.fromisoformat(end_time_str)
            now = self._vtherm.now
            
            # 3. Re-populate internal manager variables
            self._is_timed_preset_active = True
            self._timed_preset = VThermPreset(preset_str)
            self._timed_preset_end_time = end_time

            # 4. Reschedule the expiration task if time remains
            if end_time > now:
                _LOGGER.info("%s - Resuming timed preset %s. Reverting at %s", self, preset_str, end_time)
                self._cancel_timer = async_track_point_in_time(
                    self._hass,
                    self._async_timed_preset_expired,
                    self._timed_preset_end_time,
                )
            else:
                _LOGGER.info("%s - Timed preset expired during downtime. Cleanup will follow.", self)
                # The existing safety check in refresh_state() will handle the cleanup 
                # and revert to requested_state during the startup cycle.
[...]

Issue #1654, opened 2026-01-20
At startup, VTherm in auto stop mode stop can be not in sync with underlying Heat Pump Split.


jcmaison:
## Version of the custom_component

Versatile Thermostat 8.6.0

## Configuration

vTherm PAC Lucas configuration:
```
hvac_modes:
  - "off"
  - heat
  - dry
  - cool
  - fan_only
min_temp: 10
max_temp: 30
target_temp_step: 0.5
fan_modes:
  - auto
  - "1"
  - "2"
  - "3"
  - "4"
  - "5"
preset_modes:
  - none
  - frost
  - eco
  - comfort
  - boost
swing_modes:
  - auto
  - 1_up
  - "2"
  - "3"
  - "4"
  - 5_down
  - swing
swing_horizontal_modes:
  - auto
  - 1_left
  - "2"
  - "3"
  - "4"
  - 5_right
  - split
  - swing
current_temperature: 17.8
temperature: 10
fan_mode: "1"
hvac_action: idle
preset_mode: frost
swing_mode: auto
swing_horizontal_mode: "3"
hvac_mode: "off"
ema_temp: 17.91
specific_states:
  is_on: false
  last_central_mode: Heat only
  last_update_datetime: "2026-01-20T13:24:30.323348+01:00"
  ext_current_temperature: 14.1
  last_temperature_datetime: "2026-01-20T13:18:57.351051+01:00"
  last_ext_temperature_datetime: "2026-01-20T13:04:46.190537+01:00"
  is_device_active: false
  device_actives: []
  nb_device_actives: 0
  ema_temp: 17.91
  temperature_slope: -0.46
  hvac_off_reason: hvac_off_auto_start_stop
  total_energy: 0
  last_change_time_from_vtherm: "2026-01-20T08:20:24.017129+01:00"
  messages:
    - hvac_off_auto_start_stop
  is_sleeping: false
  is_locked: false
  is_recalculate_scheduled: false
configuration:
  ac_mode: true
  type: over_climate
  is_controlled_by_central_mode: true
  target_temperature_step: 0.5
  timezone: Europe/Paris
  temperature_unit: °C
  is_used_by_central_boiler: false
  max_on_percent: null
  have_valve_regulation: false
  cycle_min: 5
preset_temperatures:
  frost_temp: 10
  eco_temp: 19
  boost_temp: 21
  comfort_temp: 20
  frost_away_temp: 10
  eco_away_temp: 17
[...]

PR #1917, opened 2026-04-01
Fix Auto Start Stop Status

adamotte:
# Fix: auto_start_stop `hvac_off_reason` not restored after HA restart

### Bug

When a VTherm `over_climate` is stopped by the **auto start/stop** feature (`hvac_off_reason: hvac_off_auto_start_stop`), a Home Assistant restart causes the VTherm to switch back to `hvac_mode: heat` while the underlying device correctly remains `off`.

This is the **inverse** of issue #1654 (where the underlying was not synced): here the VTherm itself "forgets" it was auto-stopped and resumes heating, potentially sending unwanted commands to the underlying device later.

### Root cause

`BaseThermostat.get_my_previous_state()` iterates over all managers and calls `manager.restore_state(old_state)` (see base_thermostat.py line ~475).

`FeatureAutoStartStopManager` does **not** implement `restore_state()`. Consequently, after state restoration:

- `_hvac_off_reason` is reset to `None` instead of `HVAC_OFF_REASON_AUTO_START_STOP`
- `_saved_hvac_mode` is reset to `None` instead of the pre-stop mode (e.g. `heat`)

The VTherm then sees `hvac_off_reason=None` and `hvac_mode=off`, concludes there is no reason to stay off, and transitions back to `heat`.

### Fix

Add a `restore_state()` method to `FeatureAutoStartStopManager` that:
1. Reads `hvac_off_reason` from the persisted entity attributes.
2. If the reason is `hvac_off_auto_start_stop`, restores both `_hvac_off_reason`    and `_saved_hvac_mode` on the parent VTherm so the auto-stop state survives  the restart.

This follows the exact same pattern used by `FeatureWindowManager` and the recently added `FeatureTimedPresetManager` fix (#1722).

## Changes

| File | Change |
|------|--------|
| `custom_components/versatile_thermostat/feature_auto_start_stop_manager.py` | Add `restore_state()` method |
| `tests/test_auto_start_stop_restore.py` | Add unit test covering restart scenario |

## How to test

1. Configure a VTherm `over_climate` with auto start/stop enabled.
2. Let the auto start/stop trigger (VTherm switches to `off`,
   `hvac_off_reason` = `hvac_off_auto_start_stop`).
3. Restart Home Assistant.
4. **Before fix**: VTherm restarts in `heat` mode, underlying stays `off`.  
   **After fix**: VTherm remains `off`, `hvac_off_reason` is preserved.

## Related issues

- Related to #1654 (inverse symptom, different root cause)
- Same pattern as #1722 (TimedPresetManager missing restore_state)

## Misc

This PR has been created with the help of Anthropic Claude Sonnet 4.6


custom_components/versatile_thermostat/feature_auto_start_stop_manager.py
jmcollin78, line 35: Because thoses attributes are not saved, I wonder how it could work.
I understand you restore the state but if you don't restore the calculation results if could not work as expected.

jmcollin78:
In my mind, auto-start/stop shouldn't be saved and restored. It was done on purpose. Suppose there is a bug and the device don't restart. A restart of HA should restore the correct behavior.

This is also the case for other managers like window manager. 

But maybe I'm wrong and the auto-start/stop should be restored because with HA we need to restart often due to update. Or I need a reset state action in case of Vtherm being stuck.

adamotte:
> In my mind, auto-start/stop shouldn't be saved and restored. It was done on purpose. Suppose there is a bug and the device don't restart. A restart of HA should restore the correct behavior.

For my point of view Versatile Thermostat is not responsible of a bug in the underlying devices. It should restore its own state properly.

I have in mind a problem with DAIKIN climate (that I own) which don't respond to HVAC mode only, you need fo force a `climate.turn_on`. I'm OK with the fact that I had to handle it myself with an automation.

> But maybe I'm wrong and the auto-start/stop should be restored because with HA we need to restart often due to update. Or I need a reset state action in case of Vtherm being stuck.

That's exactly my use case. With your integration, or others and even with HA himself, I have to update and so restart my instace at least 2 times per month. Those days, my climate are screw until I synchronized themself manually. Normally I let them handle themself between auto start stop feature and automation for target temperature.

The second problem behind is that as the VTerm is trying to heat with no success (because underlying is stopped), calculated temperatures, slope or errors could be degraded (as per my understanding)

jmcollin78:
## Code Review
Thank you for this contribution. After carefully analyzing the PR against the actual codebase, I have identified several significant issues that prevent merging it as-is.

1. The described bug does not actually exist (for [hvac_off_reason](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html))
In [base_thermostat.py](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), [hvac_off_reason](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is already restored before any manager's [restore_state()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is called:

So [hvac_off_reason](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is already correctly restored through the base class — no manager override is needed for this.

2. saved_hvac_mode is not a VTherm attribute
The patch assigns:

saved_hvac_mode does not exist as a property or attribute on [BaseThermostat](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). It only appears in event payloads (inside [send_event()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) data dicts). This creates a dynamic Python attribute that is never read — dead code.

3. Reading data that is never persisted
The patch reads:

Looking at [add_custom_attributes()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), the [auto_start_stop_manager](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) dict stored in entity state contains: [auto_start_stop_enable](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), [auto_start_stop_level](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), auto_start_stop_dtmin, [is_auto_stop_detected](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), etc. — but never saved_hvac_mode. This key is only sent in events and never stored in entity attributes, so this read will always return None.

4. The guard condition is counterproductive
This would silently skip restoration for states saved with older VTherm versions that didn't yet store this dict — making the fix version-dependent in the wrong direction.

5. Tests validate a mock, not the real object
Since MagicMock accepts any attribute assignment without error, the tests pass successfully even though saved_hvac_mode doesn't exist on the real VTherm. The integration test is also explicitly pytest.skip()-ped.

What might actually need fixing
If there is a real restart-related issue, it would more likely be [_is_auto_stop_detected](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (the algorithm's internal boolean) not being restored. After a restart it defaults to False. If the algorithm recalculates and finds [should_be_off = True](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), it detects a "change" (False → True) and triggers [update_states()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) unnecessarily.

The correct fix would be to restore this flag in [restore_state()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) from the already-persisted [is_auto_stop_detected](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) key inside the [auto_start_stop_manager](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) attribute dict:

Conclusion
This PR should not be merged in its current state. The described root cause (line ~475) and the proposed fix address a problem that the base class already handles. I recommend closing it and, if the restart behavior is genuinely problematic, opening a new PR focused solely on restoring [_is_auto_stop_detected](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) with proper integration tests following the existing patterns in the test suite.

jmcollin78:
I will close and do what is suggested: only restore the `is_auto_stop_detected`.

I will do my own PR. Thank you anyway for this nice try and for highlighting this issue.

PR #1922, opened 2026-04-02
Fix valve cycle completion callbacks in cycle scheduler

KipK:
End cycle callback was missing for valve mode in scheduler.

Added cycle_duration_min feedback to algorithm ( not used by Auto-TPI yet  but will be, already used on SmartPI )


jmcollin78 (approved):
Je ne suis pas sur de bien comprendre le sujet, mais je te fais confiance

Issue #1671, opened 2026-01-22
[Feature request] - Allow the user to specify the unit of all power attributes and adapt to unit of the power sensors


superdarco78:
Rejestrator: homeassistant.components.sensor
Źródło: components/sensor/__init__.py:729
integracja: Sensor (dokumentacja, Problemy)
Pierwsze zdarzenie: 17:21:31 (1 wystąpienie)
Ostatnio zalogowany: 17:21:31

Entity sensor.total_power_active_for_boiler (<class 'custom_components.versatile_thermostat.sensor.TotalPowerActiveDeviceForBoilerSensor'>) is using native unit of measurement 'None' which is not a valid unit for the device class ('power') it is using; expected one of ['MW', 'kW', 'W', 'mW', 'GW', 'TW']; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/jmcollin78/versatile_thermostat/issues

jmcollin78:
Hello, a preview is available here if you want to test it https://github.com/jmcollin78/versatile_thermostat/releases/tag/10.3.0.beta1
Note : i will do some tests at home in parallel.

jmcollin78:
https://github.com/jmcollin78/versatile_thermostat/releases/tag/10.3.0

PR #1931, opened 2026-04-07
Fix false power shedding on VTherm startup with multiple underlyings

KipK:
The power shedding logic could incorrectly trigger when a single VTherm controls multiple underlyings.

Root cause: during a startup sequence, each underlying turn_on() path called the temporary startup power reservation logic, but the reserved value was computed at VTherm scope. In a multi-underlying setup, if several underlyings started before Home Assistant had reflected the first ON state, the same VTherm startup power could be reserved multiple times.

This led to:
- false "not enough power available" detection when starting a VTherm with multiple underlyings,
- erroneous transition to power shedding even though the meter still had enough available power,
- possible lingering shedding afterwards until the next central power sensor update recalculated the state.

Changes:
- replace the temporary started power accumulator with a per-VTherm reservation map in the central power manager,
- make startup power reservation idempotent per VTherm during a sensor update window,
- make startup power release idempotent as well,
- keep the central behavior of clearing temporary reservations on the next power sensor change.

Tests:
- add a regression test proving that a multi-underlying VTherm reserves its startup power only once even if several underlyings start in the same startup sequence before HA state catches up,
- keep existing power management and multi-switch power tests passing.

This fixes the main false shedding scenario observed on startup for multi-underlying VTherms, without changing the overall shedding strategy.

### DIFF
--- a/custom_components/versatile_thermostat/feature_central_power_manager.py
+++ b/custom_components/versatile_thermostat/feature_central_power_manager.py
@@ -50,7 +50,7 @@ class FeatureCentralPowerManager(BaseFeatureManager):
         self._current_max_power: float | None = None
         self._power_temp: float | None = None
         self._cancel_calculate_shedding_call = None
-        self._started_vtherm_total_power: float = 0
+        self._started_vtherm_total_power_by_id: dict[str, float] = {}
         # Not used now
         self._last_shedding_date = None
         self._state = False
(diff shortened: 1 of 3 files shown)


jmcollin78 (approved):
Je vois l'idée. En effet, on ne serait plus très loin d'avoir une puissance différente par under. On vire la puissance de l'éqt et on la met dans chaque under. Et on en profite pour faire https://github.com/jmcollin78/versatile_thermostat/issues/1671. Ca me parait un bon plan pour une prochaine version.

KipK:
> Je vois l'idée. En effet, on ne serait plus très loin d'avoir une puissance différente par under. On vire la puissance de l'éqt et on la met dans chaque under. 

Yes c'est l'idée de préparer pour ça. Et ça permettra un délestage plus granulaire ( i.e n'utiliser que 2 underlyings à la fois sur les 4 par ex, mais je pense qu'il faudra rajouter une petite chose côté scheduler pour une prise en charge correcte )

PR #1936, opened 2026-04-09
Fix startup power reservation for multi-underlying VTherms

KipK:
Use per-underlying startup reservation instead of a single per-VTherm reservation during the power-sensor stale window.

Changes:
- reserve `device_power / nb_underlying_entities` for each underlying startup on multi-switch VTherms
- stop using `device_power * on_percent` for startup authorization
- keep idempotence for repeated reservation of the same underlying
- store temporary reservations by underlying `entity_id`
- remove the unused legacy reservation path
- update regression tests for per-underlying reservation semantics

This fixes the case where a multi-switch VTherm could start several underlyings even though only one underlying power slice was actually available.

### DIFF
--- a/custom_components/versatile_thermostat/feature_central_power_manager.py
+++ b/custom_components/versatile_thermostat/feature_central_power_manager.py
@@ -295,26 +295,21 @@ class FeatureCentralPowerManager(BaseFeatureManager):
         vtherms.sort(key=cmp_to_key(cmp_temps))
         return vtherms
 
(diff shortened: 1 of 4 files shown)


jmcollin78 (approved):
Thank you. Will do more test this week-end.

KipK:
Yes c'est a tester qd même. J'ai plus de chauffage en route depuis 10j et pas de banc de test opérationnel pour ça.

jmcollin78:
Je suis en train de tester depuis hier. Ca se passe bien.

PR #1946, opened 2026-04-18
fix cycle scheduler elapsed time for 0 and 100 percent cycles

KipK:


KipK:
no need to backport to 9.x , only smartpi use this yet.

jmcollin78 (approved):

PR #1949, opened 2026-04-21
Notify external prop handlers on every thermostat state refresh

KipK:
External proportional plugins may need to react to temperature crossings even when VT logical state did not change.


KipK:
oups wrong branch

PR #1950, opened 2026-04-21
Notify external prop handlers on every thermostat state refresh

KipK:
External proportional plugins may need to react to temperature crossings even when VT logical state did not change.

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_prop.py
+++ b/custom_components/versatile_thermostat/thermostat_prop.py
@@ -242,7 +242,9 @@ class ThermostatProp(BaseThermostat[T], Generic[T]):
     async def update_states(self, force=False):
         """Update states and delegate to handler."""
         changed = await super().update_states(force)
-        if changed and self._algo_handler:
+        if self._algo_handler:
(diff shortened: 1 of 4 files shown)


custom_components/versatile_thermostat/thermostat_prop.py
```diff
@@ -242,7 +242,9 @@ async def _control_heating_specific(self, timestamp=None, force=False):
     async def update_states(self, force=False):
         """Update states and delegate to handler."""
         changed = await super().update_states(force)
-        if changed and self._algo_handler:
+        if self._algo_handler:
+            # External proportional plugins may need to react to temperature
+            # crossings even when VT logical state did not change.
             await self._algo_handler.on_state_changed()
```
jmcollin78: Give changed to the on_state_changed() for future use and ascendant compatibility.

jmcollin78 (changes requested):

KipK:
Du coup je t'ai mis le PR pour l'api aussi. 

https://github.com/jmcollin78/vtherm_api/pull/3

jmcollin78 (approved):

PR #1967, opened 2026-05-13
External Humidity Sensor

BTopbas:
Adds optional `humidity_sensor_entity_id` config field to Versatile Thermostat.

When set, `current_humidity` is sourced from the external sensor instead of the underlying climate entity. This makes the value appear natively in the Lovelace thermostat card and Apple Home.

Useful when the underlying entity either reports humidity from a non-representative location, or does not report humidity at all.

No breaking changes — existing behavior is fully preserved when the field is not configured.


jmcollin78 (commented):
Seems good for me. But please add the humidity value into the custom-attributes (see update_custom_attributes). Put it into the `specific_states` structure.

You can also proceed to translate to the other languages. (IA does that very well for you)

Thank you !

jmcollin78:
Please rebase to get the new chinese translation.

jmcollin78:
Hello, any news on this PR ?

BTopbas:
> Hello, any news on this PR ?

I'm currently on vacation and away from my computer. I will make these minor fixes as soon as I'm back."

KipK:
@jmcollin78  HA climate entity has current_humiity attribute in the specs. It should be there instead of a specific_states one.

jmcollin78:
No new since May. I close. I will do it myself.

PR #1973, opened 2026-05-22
Update proportional algorithm descriptions with plugin docs links

KipK:
Change the deprecated description in underlyings

image


jmcollin78 (approved):
Thank you !

PR #1981, opened 2026-05-27
Register underlying entities before startup state reads

rosenrot00:
#### Summary
- Register underlying entity IDs in `UnderlyingStateManager` when the underlying entity is created.
- Refresh registered states when startup listeners are attached.
- Keep the error log for truly unmanaged entity IDs, with managed IDs included for diagnostics.
- Add tests for pre-startup state reads and startup refresh behavior.

#### Rationale
During startup, VTherm can request the state of a configured underlying entity before `add_underlying_entities()` has attached listeners. Those reads were logged as unknown entity IDs even though the entities were valid. Registering the IDs earlier fixes the lifecycle ordering instead of hiding the error.

### DIFF
--- a/custom_components/versatile_thermostat/underlying_state_manager.py
+++ b/custom_components/versatile_thermostat/underlying_state_manager.py
@@ -78,7 +78,11 @@ class UnderlyingStateManager:
         """Return the last known `State` for `entity_id`, or `None` if unknown."""
         idx = self._index_of(entity_id)
         if idx is None:
-            _LOGGER.error("UnderlyingStateManager - Requested state for unknown entity_id: %s", entity_id)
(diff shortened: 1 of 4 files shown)


jmcollin78:
Hello, thank you for this PR.

The goal of all the underlying_state_manager is to register an entity when its state has been received. So it is normal to have the `UnderlyingStateManager - Requested state for unknown entity_id` if the state takes time to come.

Changing this logic is impacting because it breaks the goal. We should not register an entity which doesn't have a state. That was the v8 behavior and this leads to many issues.

So if i understand correctly your PR, I will not integrate it. And maybe the only thing to do is to change the log level from error to info (or debug because this is a temporary message most of the time).

PR #1982, opened 2026-05-27
Add humidity regulation for climate thermostats

rosenrot00:
#### Summary
- Add an optional humidity regulation feature for `thermostat_over_climate`.
- Support an external humidity sensor, with fallback to the underlying climate humidity when available.
- Expose target humidity through the standard Home Assistant climate API.
- Add dry-only humidity regulation and cool-mode humidity assist.
- Add humidity hysteresis/tolerance to avoid short cycling.
- Add translations and tests for the humidity behavior.

#### Behavior
In dry-only mode, VTherm regulates humidity only while the thermostat is in `dry` mode. It starts dehumidification when current humidity is above target plus tolerance, and stops when the target humidity is reached.

In cool-assist mode, normal cooling remains temperature-driven. Once the temperature target is satisfied, VTherm can temporarily switch to dry mode if humidity is still too high.


rosenrot00:
@jmcollin78 I hope this PR is in your interest as well. Please give me some more iterations, I will test it on my end.

Tested and works well.

jmcollin78:
Hello, this is an interesting feature addition. But the goal of VTherm is to regulate the temperature. It is too muche changes for a small addition. I don't know if someone regulated the humidity through a thermostat.

But I have now the solution: you should do a plugin and add the humidityManager as a new plugable feature. This is now possible since 10.0.x release. You have the list of existing plugin here: https://www.versatile-thermostat.org/en/plugins/
and the roadmap to add plugin is described here: https://github.com/jmcollin78/vtherm_api/tree/main/documentation/en

A plugin for a Manager has never been used until now but should be possible because it is on the initial specifications.

Of course I can help if you need.

I close but we can continue to discuss of course

rosenrot00:
Thanks for the feedback. I agree that the PR grew larger than a small addition, and moving the humidity manager into a pluggable feature sounds reasonable if that is the preferred architecture for VTherm 10.x.

That said, I think humidity regulation is not completely outside the climate/thermostat domain. Home Assistant’s ClimateEntity model explicitly supports current_humidity, target_humidity, and TARGET_HUMIDITY, and climate.set_humidity is a standard climate service. Some underlying climate devices also expose humidity control directly, especially AC/dehumidification devices.

So my goal was not to turn VTherm into a generic humidifier controller, but to let thermostat_over_climate preserve and regulate humidity behavior already available on the underlying climate device, especially in dry mode.

If the plugin approach is the right direction, I’m happy to rework it that way.

PR #1988, opened 2026-05-29
Fix missing unit on total_power_active_for_boiler sensor

dark-light-cz:
TotalPowerActiveDeviceForBoilerSensor declares device_class power but did not expose native_unit_of_measurement, triggering HA startup warnings.


dark-light-cz:
Please take a look at this. It's just small fix of startup warning.

jmcollin78:
Hello, thank you for your PR. I understand your fix but there is something you don't understand: the unit has no matter on VTherm. What is important is that all power unit are the same watt or kw or Mw .. Vtherm don't care.

But to fix the warning, I intend to ask the user what is the real power unit. See: #1671 . This is the real fix.

If you have time, this could be a very interesing and more useful PR to do.

PR #2004, opened 2026-06-19
Fix #2003: set_temperature crashes when underlying supported_features is a plain int

### LINKED ISSUE #2003: over_climate: set_temperature() crashes with "TypeError: argument of type 'int' is not a container or iterable" when the underlying's supported_features is a plain int
## Summary

For an `over_climate` VTherm, `UnderlyingClimate.set_temperature()` can crash with:

```
TypeError: argument of type 'int' is not a container or iterable
```

every time VTherm pushes a setpoint to the underlying `climate` entity. It is **intermittent** and "fixes itself after a few tries", which makes it look transient — but it is a deterministic type bug, not a race in the usual sense.

The crash happens because the membership operator `in` is used against `self.supported_features`, and that value is read from the underlying entity's **state attributes**, where Home Assistant can store `supported_features` as a **plain `int`** instead of a `ClimateEntityFeature` `IntFlag`. `EnumMember in <int>` raises, because a plain `int` is not iterable.

## Affected version

- Reproduced on **10.0.2** (latest release).
- Still present on **`main`** at the time of writing — both offending lines are unchanged.

## Where (the bug)

`custom_components/versatile_thermostat/underlyings.py`, in `UnderlyingClimate.set_temperature()`:

```python
# Issue 807 add TARGET_TEMPERATURE only if in the features
if ClimateEntityFeature.TARGET_TEMPERATURE_RANGE in self.supported_features:   # <-- line 864
    data.update(
        {
            "target_temp_high": target_temp,
            "target_temp_low": target_temp,
        }
    )

if ClimateEntityFeature.TARGET_TEMPERATURE in self.supported_features:          # <-- line 872
    data["temperature"] = target_temp
```

`self.supported_features` is defined a few lines below as:

```python
@property
def supported_features(self) -> ClimateEntityFeature:
    """Get the supported features of the climate"""
    return self.get_underlying_attribute("supported_features") or ClimateEntityFeature(0)
```

and `get_underlying_attribute` returns the raw value straight out of the HA state object:

```python
def get_underlying_attribute(self, attribute_name: str) -> Any:
    """Get an attribute from the underlying climate"""
    if not self.is_initialized:
        return None
    state = self._state_manager.get_state(self._entity_id)
    if state is None:
        return None
    return state.attributes.get(attribute_name, None)
```

## Root cause — why `in` is wrong here, and why it's intermittent

`ClimateEntityFeature` is an `IntFlag`. The `in` operator behaves very differently depending on the **runtime type** of the right-hand operand:

- `ClimateEntityFeature.TARGET_TEMPERATURE in ClimateEntityFeature(...)` → **valid** (IntFlag supports membership tests).
- `ClimateEntityFeature.TARGET_TEMPERATURE in 393` → **`TypeError: argument of type 'int' is not a container or iterable`** (a plain `int` is not iterable, and `in` falls back to iteration).

`self.supported_features` therefore works or throws depending on what type the underlying entity's state happens to be carrying for the `supported_features` attribute **at that instant**:

| Underlying state at call time | `supported_features` attr type | `X in supported_features` |
|---|---|---|
| Live state from a healthy entity | `ClimateEntityFeature` (IntFlag) | ✅ works |
| **Restored / rehydrated state** (entity just came back from `unavailable`, or its integration was just reloaded) | **plain `int`** (deserialized from JSON) | ❌ **TypeError** |
| Attribute missing / entity not yet initialized | falls back to `ClimateEntityFeature(0)` | ✅ works |

So the bug fires specifically in the short window right after the underlying climate entity transitions through `unavailable` or its integration is reloaded — exactly when the state has been restored from the cache (where attributes are plain JSON ints) and the live coordinator update hasn't replaced it yet. A second or two later the live state lands, `supported_features` is an `IntFlag` again, and the retry succeeds. Hence "it goes away after a few tries."

This is also **inconsistent with the rest of the same class**, which already does the correct thing with a bitwise test:

```python
# fan_modes / swing_modes / swing_horizontal_modes properties
if not self.is_initialized or self.supported_features & ClimateEntityFeature.FAN_MODE == 0:
    ...
if not self.is_initialized or self.supported_features & ClimateEntityFeature.SWING_MODE == 0:
    ...
if not self.is_initialized or self.supported_features & ClimateEntityFeature.SWING_HORIZONTAL_MODE == 0:
    ...
```

The bitwise `&` works for **both** a plain `int` and a `ClimateEntityFeature`. Only the two `in` checks added for "Issue 807" use the fragile membership form.

## Full traceback (real, from an over_climate with self-regulation)

```
File "/usr/src/homeassistant/homeassistant/components/climate/__init__.py", line 630, in async_handle_set_preset_mode_service
    await self.async_set_preset_mode(preset_mode)
File "/config/custom_components/versatile_thermostat/base_thermostat.py", line 1371, in wrapper
File "/config/custom_components/versatile_thermostat/base_thermostat.py", line 1418, in async_set_preset_mode
File "/config/custom_components/versatile_thermostat/base_thermostat.py", line 1548, in update_states
File "/config/custom_components/versatile_thermostat/thermostat_climate.py", line 840, in async_control_heating
File "/config/custom_components/versatile_thermostat/thermostat_climate.py", line 260, in _send_regulated_temperature
File "/config/custom_components/versatile_thermostat/underlyings.py", line 864, in set_temperature
TypeError: argument of type 'int' is not a container or iterable
```

The same crash is reachable from any path that calls `under.set_temperature(...)` (manual temperature change, preset change, or a periodic auto-regulation cycle) — it is not specific to `set_preset_mode`. Auto-regulation just makes it far more likely to be observed, because it pushes the setpoint on every cycle, multiplying the number of chances to land inside the "restored state" window.

## Real-world impact example

**Setup:** An `over_climate` VTherm controlling an **LG multi-split heat pump (VRF) indoor unit** via the official `lg_thinq` cloud integration. Self-regulation is enabled (Light), so VTherm sends a (compensated) setpoint to the LG unit on every regulation cycle.

`lg_thinq` is a cloud integration and its entities periodically flap to `unavailable` on transient cloud hiccups (`NOT_CONNECTED_DEVICE`, `RETRY_REQUEST`, request timeouts), and the integration is sometimes reloaded to recover control. Each `unavailable → available` transition and each reload opens a brief window where the LG climate entity's state is a **restored** state whose `supported_features` is a plain `int`.

**Symptom the user sees:** dragging the temperature dial (or tapping a preset) on the VTherm in the HA app fails with the toast:

> Failed to perform the action climate/set_preset_mode. argument of type 'int' is not a container or iterable

Retrying a few seconds later succeeds. Beyond the annoying UI error, **a regulation cycle that lands in this window is silently lost** — the computed setpoint is never delivered to the heat pump, so the room can drift from target until the next successful cycle. On a flaky cloud integration this can recur many times a day.

This is a great example of the bug because (a) it's a genuinely common hardware/integration combo, (b) the underlying flakiness guarantees frequent restored-state windows, and (c) it shows the bug has real control consequences, not just a cosmetic toast.

## Reproduction (minimal)

1. Create an `over_climate` VTherm over any underlying `climate` entity.
2. Force the underlying entity's state to carry `supported_features` as a **plain int** (this is what HA does for a restored state — e.g. set the state attribute to `int(ClimateEntityFeature.TARGET_TEMPERATURE)` instead of the `IntFlag`).
3. Trigger a setpoint push (change temperature / preset, or wait for an auto-regulation cycle).
4. Observe the `TypeError`.

## Proposed fix

Replace the two `in` membership checks with the bitwise test already used everywhere else in the class:

```python
if self.supported_features & ClimateEntityFeature.TARGET_TEMPERATURE_RANGE:
    ...

if self.supported_features & ClimateEntityFeature.TARGET_TEMPERATURE:
    ...
```

This is correct for both a plain `int` and a `ClimateEntityFeature` `IntFlag`, and keeps the file internally consistent.

PR with the fix + a regression test follows.

jasonjhofmann:
### Problem

For an `over_climate` VTherm, `UnderlyingClimate.set_temperature()` can crash with:

```
TypeError: argument of type 'int' is not a container or iterable
```

every time VTherm pushes a setpoint to the underlying `climate` entity (manual temperature change, preset change, or an auto-regulation cycle). It is intermittent and "goes away after a few tries", but it is a deterministic type bug.

`set_temperature()` used the membership operator `in` against `self.supported_features`:

```python
if ClimateEntityFeature.TARGET_TEMPERATURE_RANGE in self.supported_features:
    ...
if ClimateEntityFeature.TARGET_TEMPERATURE in self.supported_features:
    ...
```

`self.supported_features` is read from the underlying entity's **state attributes**:

```python
return self.get_underlying_attribute("supported_features") or ClimateEntityFeature(0)
```

Home Assistant stores `supported_features` as a **plain `int`** for restored states (e.g. right after the underlying entity returns from `unavailable`, or its integration is reloaded). `ClimateEntityFeature.X in <int>` raises, because a plain `int` is not iterable. When the live state (carrying the `ClimateEntityFeature` `IntFlag`) lands a moment later, the retry succeeds — hence the intermittency.

A worse consequence than the UI toast: an auto-regulation cycle that lands in that window is silently dropped, so the underlying device never receives the computed setpoint until the next successful cycle. See #2003 for a full real-world example (LG VRF heat pump via the `lg_thinq` cloud integration, whose entities frequently flap to `unavailable`).

### Fix

Use a bitwise test instead of `in`. This is correct for both a plain `int` and a `ClimateEntityFeature` `IntFlag`, and matches the pattern already used elsewhere in the same class (`fan_modes` / `swing_modes` / `swing_horizontal_modes`):

```python
if self.supported_features & ClimateEntityFeature.TARGET_TEMPERATURE_RANGE:
    ...
if self.supported_features & ClimateEntityFeature.TARGET_TEMPERATURE:
    ...
```

### Test

Adds `test_bug_2003` to `tests/test_bugs.py`, parametrized over `supported_features` being a **plain `int`** (the restored-state case that previously crashed) and a **`ClimateEntityFeature` IntFlag** (the healthy case). It asserts `set_temperature()` does not raise and that the underlying `climate.set_temperature` service call is issued with the correct payload. The test fails on `main` (TypeError for the plain-int parametrization) and passes with this fix.

### Checklist

- [x] Code lints with `black`
- [x] Regression test added and passing
- [x] No behavior change for the healthy (IntFlag) path

### DIFF
--- a/custom_components/versatile_thermostat/underlyings.py
+++ b/custom_components/versatile_thermostat/underlyings.py
@@ -860,8 +860,15 @@ class UnderlyingClimate(UnderlyingEntity):
 
         _LOGGER.info("%s - Set setpoint temperature to: %s", self, target_temp)
 
-        # Issue 807 add TARGET_TEMPERATURE only if in the features
-        if ClimateEntityFeature.TARGET_TEMPERATURE_RANGE in self.supported_features:
+        # Issue 807 add TARGET_TEMPERATURE only if in the features.
+        # Use a bitwise test (not the `in` operator): self.supported_features is read
+        # from the underlying entity's state attributes, where Home Assistant may store
+        # it as a plain int (e.g. for restored states). `EnumMember in <int>` raises
+        # `TypeError: argument of type 'int' is not a container or iterable`, whereas
+        # the bitwise `&` works for both a plain int and a ClimateEntityFeature IntFlag.
+        # This matches the bitwise pattern already used elsewhere in this class
+        # (fan_modes / swing_modes).
+        if self.supported_features & ClimateEntityFeature.TARGET_TEMPERATURE_RANGE:
             data.update(
                 {
                     "target_temp_high": target_temp,
@@ -869,7 +876,7 @@ class UnderlyingClimate(UnderlyingEntity):
                 }
             )
 
-        if ClimateEntityFeature.TARGET_TEMPERATURE in self.supported_features:
+        if self.supported_features & ClimateEntityFeature.TARGET_TEMPERATURE:
             data["temperature"] = target_temp
 
         try:
--- a/tests/test_bugs.py
+++ b/tests/test_bugs.py
@@ -1340,3 +1340,66 @@ async def test_bug_1947(hass: HomeAssistant, skip_hass_states_is_state):
     assert vtherm.hvac_off_reason is None
     assert fake_underlying_climate.hvac_mode == VThermHvacMode_HEAT
     assert fake_underlying_climate.target_temperature == 21.0
+
+
+@pytest.mark.parametrize(
+    "supported_features",
+    [
+        # The bug: a restored / rehydrated underlying state carries
+        # `supported_features` as a plain int (this is how Home Assistant stores
+        # it for restored states). `EnumMember in <int>` used to raise TypeError.
+        int(ClimateEntityFeature.TARGET_TEMPERATURE | ClimateEntityFeature.FAN_MODE),
+        # The healthy case: a live state carries the ClimateEntityFeature IntFlag.
+        ClimateEntityFeature.TARGET_TEMPERATURE | ClimateEntityFeature.FAN_MODE,
+    ],
+    ids=["plain_int_restored_state", "intflag_live_state"],
+)
+async def test_bug_2003(hass: HomeAssistant, supported_features):
+    """Bug #2003 - UnderlyingClimate.set_temperature() must not crash when the
+    underlying entity's `supported_features` state attribute is a plain int.
+
+    Home Assistant stores `supported_features` as a plain int for restored states
+    (e.g. right after the underlying entity comes back from `unavailable` or its
+    integration is reloaded). The previous code used
+    `ClimateEntityFeature.X in self.supported_features`, which raises
+    `TypeError: argument of type 'int' is not a container or iterable` for a plain
+    int. The fix uses a bitwise test, which works for both a plain int and a
+    ClimateEntityFeature IntFlag.
+    """
+
+    thermostat = MagicMock()
+
+    under = UnderlyingClimate(hass=hass, thermostat=thermostat, climate_entity_id="climate.test")
+
+    # Reproduce the underlying state: `supported_features` of the parametrized type,
+    # plus min/max so clamp_sent_value() works.
+    under_state = State(
+        "climate.test",
+        HVACMode.COOL,
+        attributes={
+            "supported_features": supported_features,
+            "min_temp": 7.0,
+            "max_temp": 35.0,
+            "temperature": 20.0,
+        },
+    )
+    under.state_manager.get_state = MagicMock(return_value=under_state)
+
+    # Capture the underlying service call instead of really issuing it.
+    under.hass_services_async_call = AsyncMock()
+
+    with patch.object(type(under), "is_initialized", new_callable=PropertyMock, return_value=True):
+        # Must not raise (it used to raise TypeError for the plain-int case).
+        await under.set_temperature(20.0, 35.0, 7.0)
+
+    # The setpoint was delivered to the underlying climate entity.
+    assert under.hass_services_async_call.await_count == 1
+    call_args = under.hass_services_async_call.await_args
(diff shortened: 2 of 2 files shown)


jmcollin78 (approved):
Thank you for this fix !

PR #2005, opened 2026-06-20
Add optional fan-only action for auto-start/stop

krzkraw:
## Summary

This PR extends the existing auto-start/stop feature for `over_climate` thermostats with an optional `fan_only` auto-stop action.

The current behavior remains the default:

* `Turn off` — existing behavior, switches VTherm and the underlying climate entity to `off`
* `Fan only` — switches VTherm and the underlying climate entity to `fan_only` while auto-stop is active, when supported

If `fan_only` is selected but the underlying climate entity does not support it, VTherm falls back to `off`.

## Motivation

I use VTherm with many devices and really like the auto-start/stop feature. My motivation here was to make it less intrusive for bedrooms and split AC units, where fully turning the unit off and back on can cause noise, beeps, compressor cycling, or airflow changes during the night.

For those cases, `fan_only` can stop active cooling/heating while keeping the unit in a quieter and less disruptive state.

## Implementation notes

The goal was to keep this as a small extension of the current feature:

* auto-start/stop detection logic is unchanged
* default behavior remains `turn_off`
* `requested_state` remains the requested heating/cooling mode, so auto-start can restore it
* `current_state` temporarily becomes `fan_only` when selected and supported
* temperature self-regulation and auto-fan adjustments are not sent while `fan_only` auto-stop is active

image

The PR is split into three commits:

1. Functional implementation and tests
2. English UI strings and documentation
3. Additional translations

The third commit can be adjusted or dropped if maintainers prefer translations to be handled separately.

## Tests

Added/updated tests for:

* fan-only auto-stop when supported
* fallback to off when fan-only is unsupported
* preserving requested heat/cool mode while current mode is fan-only
* avoiding regulated temperature sends while fan-only auto-stop is active
* avoiding auto-fan changes while fan-only auto-stop is active

## Verification

Black check was run on changed Python files. It reports existing formatting differences in upstream files, so I did not apply full-file formatting to avoid unrelated diff noise and ballooning this PR to 600+ lines.

I tested this on my two AC units on my HA instance. Works beautiflly. Auto start/stop works exactly as before just with different action. Mixing off or fan only works as expected
image

## Possible future follow-up - If you like my code and this would align with project's goals 

I got quite confused by how auto start/stop actually operates without digging into docs. I was thinking of adding threshold parameter of how much overcooling or overheating user allows while speed of auto start/stop limits cycling (minimal cycle length) then. What do you think?

Thank you for this awesome project!

### DIFF
--- a/custom_components/versatile_thermostat/feature_auto_start_stop_manager.py
+++ b/custom_components/versatile_thermostat/feature_auto_start_stop_manager.py
@@ -35,6 +35,7 @@ class FeatureAutoStartStopManager(BaseFeatureManager):
             "auto_start_stop_level",
             "auto_start_stop_dtmin",
             "auto_start_stop_enable",
+            "auto_start_stop_action",
             "auto_start_stop_accumulated_error",
             "auto_start_stop_accumulated_error_threshold",
             "auto_start_stop_last_switch_date",
@@ -46,6 +47,7 @@ class FeatureAutoStartStopManager(BaseFeatureManager):
         super().__init__(vtherm, hass)
 
         self._auto_start_stop_level: str = AUTO_START_STOP_LEVEL_NONE
+        self._auto_start_stop_action: str = AUTO_START_STOP_ACTION_TURN_OFF
         self._auto_start_stop_algo: AutoStartStopDetectionAlgorithm | None = None
         self._is_configured: bool = False
         self._is_auto_start_stop_enabled: bool = False
@@ -64,9 +66,14 @@ class FeatureAutoStartStopManager(BaseFeatureManager):
                 entry_infos.get(CONF_AUTO_START_STOP_LEVEL, None)
                 or AUTO_START_STOP_LEVEL_NONE
             )
+            self._auto_start_stop_action = (
+                entry_infos.get(CONF_AUTO_START_STOP_ACTION, None)
+                or AUTO_START_STOP_ACTION_TURN_OFF
+            )
             self._is_configured = True
         else:
             self._auto_start_stop_level = AUTO_START_STOP_LEVEL_NONE
+            self._auto_start_stop_action = AUTO_START_STOP_ACTION_TURN_OFF
             self._is_configured = False
 
         # Initialize the enable flag synchronously so the manager is operational
@@ -126,7 +133,7 @@ class FeatureAutoStartStopManager(BaseFeatureManager):
                             "type": "stop",
                             "name": self.name,
                             "cause": "Auto stop conditions reached",
-                            "hvac_mode": str(VThermHvacMode_OFF),
+                            "hvac_mode": str(self.auto_stop_hvac_mode),
                             "saved_hvac_mode": str(self._vtherm.requested_state.hvac_mode),
                             "target_temperature": self._vtherm.target_temperature,
                             "current_temperature": self._vtherm.current_temperature,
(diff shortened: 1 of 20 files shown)


custom_components/versatile_thermostat/feature_auto_start_stop_manager.py
```diff
@@ -182,7 +189,7 @@ async def set_auto_start_stop_enable(self, is_enabled: bool):
             self._is_auto_start_stop_enabled = is_enabled
 
             # Send an event if the vtherm was off due to auto-start/stop and enable has been set to false
-            if not is_enabled and self._vtherm.hvac_mode == VThermHvacMode_OFF and self._vtherm.hvac_off_reason == HVAC_OFF_REASON_AUTO_START_STOP:
+            if not is_enabled and self.is_auto_stop_detected:
```
jmcollin78, line 192: I'm not sure about this change. Why this change ? Are you the behvior is the same ?
Before, I check that the vherm is stopped due to auto-start/stop. Now you check that auto-start/stop is detected. If we suppose the detection has been done but for what ever reason the VTherm is not stopped, then the behavior is not the same.

custom_components/versatile_thermostat/feature_auto_start_stop_manager.py
```diff
@@ -297,15 +305,27 @@ def auto_start_stop_level(self) -> str:
...
     @property
     def auto_start_stop_enable(self) -> bool:
         """Returns the auto_start_stop_enable"""
         return self._is_auto_start_stop_enabled
 
     @property
     def is_auto_stopped(self) -> bool:
-        """Returns the is vtherm is stopped and reason is AUTO_START_STOP"""
-        return self._vtherm.hvac_mode == VThermHvacMode_OFF and self._vtherm.hvac_off_reason == HVAC_OFF_REASON_AUTO_START_STOP
+        """Returns if vtherm is auto-stopped"""
```
jmcollin78, line 327: Same remark as above.

jmcollin78 (changes requested):
Hello, thank you for this nice PR. I wanted to do something like that to allow to switch to the dry mode instead of turning off.
For my opinion, this should not be a configuration option but a select entity so that the user will be able to choose dynamically which behavior he wants when stop conditions are met. Can you change this PR considering this remark ?

I have a doubt on one change you done.

jmcollin78:
>  got quite confused by how auto start/stop actually operates without digging into docs. I was thinking of adding threshold parameter of how much overcooling or overheating user allows while speed of auto start/stop limits cycling (minimal cycle length) then. What do you think?

Not sure to fully understand what you mean. There is already thresholds (and more) which are controlled by the level parameter.

jmcollin78:
I switch to draft while waiting for your reply.

jmcollin78:
Hello, I propose to take back the development of this feature. I will close this PR and make a new one.
Thank you anyway for the idea, it is a nice improvement.

krzkraw:
@jmcollin78  Sure mate! I did it only to help myself and wanted to share so others can use it too, doesn't matter which PR gets it to upstream :) 
What I meant by being confused is that current levels do not describe how the operation will actually change just by looking at them. That's why I was thinking out loud about using temperature thresholds in degrees being a possibly nice feature

jmcollin78:
Hello, you have a preview with the feature in this pre-release: https://github.com/jmcollin78/versatile_thermostat/releases/tag/10.1.0.beta3

If you can give it a try, it could be nice.

krzkraw:
> Hello, you have a preview with the feature in this pre-release: https://github.com/jmcollin78/versatile_thermostat/releases/tag/10.1.0.beta3
> 
> If you can give it a try, it could be nice.

I installed latest beta6 and see how it behaves in my bedroom and report!

jmcollin78:
It is working fine on my side. Can't wait to hear you

krzkraw:
Can confirm too - works like a charm. 
image
Left edge of the graph was my fork. Right edge is beta6 (albeit set to lower temperature, it toggles the same though)
Nice, thank you

PR #2010, opened 2026-06-30
Add supply temperature safety lockout

kienxux:
## Summary

Adds an optional supply-water temperature safety lockout for `over_switch` thermostats.

When configured, VTherm watches a supply temperature sensor and blocks underlying switch activation if the water temperature is unsafe:

- in `HEAT`, supply temperature at or above `supply_temperature_heat_max` trips the lockout,
- in `COOL`, supply temperature at or below `supply_temperature_cool_min` trips the lockout,
- `supply_temperature_tolerance` provides hysteresis before the lockout clears,
- unavailable/unknown supply temperature fails safe by blocking activation.

While tripped, the feature cancels the current cycle, turns off underlying switch entities, and makes `UnderlyingSwitch.turn_on()` refuse reactivation so keep-alive, repair, and cycle logic do not fight the lockout.

## Why

Hydronic floor-heating/floor-cooling systems may use an `over_switch` thermostat to control a circulation pump. A Home Assistant automation can turn that pump off, but VTherm can later turn it back on because normal room demand still exists. Keeping the lockout inside VTherm makes the protection part of the same command path as TPI, keep-alive, and repair logic.

## Documentation

Updated the English `over_switch` documentation with the new supply temperature safety fields and a note that this is software supervision only. Physical protection may still be required depending on the installation.

## Validation

- `python3 -m json.tool custom_components/versatile_thermostat/strings.json`
- `python3 -m json.tool custom_components/versatile_thermostat/translations/en.json`
- `python3 -m compileall -q custom_components/versatile_thermostat tests/test_safety.py`
- `git diff --check`
- `black --check custom_components/versatile_thermostat/feature_supply_temperature_safety_manager.py`

I also attempted to install the repository test requirements and run the focused regression test, but the local Python is 3.13 and `homeassistant==2026.3.1` requires Python `>=3.14.2`, so the test dependency installation cannot complete in this environment.

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -64,6 +64,7 @@ from .feature_power_manager import FeaturePowerManager
 from .feature_motion_manager import FeatureMotionManager
 from .feature_window_manager import FeatureWindowManager
 from .feature_safety_manager import FeatureSafetyManager
+from .feature_supply_temperature_safety_manager import FeatureSupplyTemperatureSafetyManager
 from .feature_auto_start_stop_manager import FeatureAutoStartStopManager
 from .feature_lock_manager import FeatureLockManager
 from .feature_timed_preset_manager import FeatureTimedPresetManager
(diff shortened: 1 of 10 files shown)


jmcollin78:
Hello, this kind of addition should be supported through a VTherm plugin and not directly in the core of the VTherm integration.

You have example with https://github.com/jmcollin78/vtherm_pellet_stove and https://github.com/jmcollin78/vtherm_api.

PR #2012, opened 2026-07-04
Improve Auto Start/Stop behavior to match HA convention

EnumC:
Thank you for making this amazing plugin! It has single handedly fixed the really bad thermo-probe readings in my Mitsubishi mini-split system. I really like the Auto Start/Stop feature, which fixed the problem with my AC never turning off the fan, even when it's not needed.

But because the current implementation turns on/off VTherm's HVACMode, it's hard to tell whether it's in heating or cooling mode when it's auto-stopped, and it's hard to distinguish an auto-stop from a user-requested off. It's also a bit confusing to set the VTherm mode from Off to Cool, and nothing happens because auto-off is on.

This PR refactors the feature to use `HVACAction` field to represent the current action of the underlying per [hass docs](https://developers.home-assistant.io/docs/core/entity/climate?_highlight=hvacaction#hvac-action).

With the refactoring, when it auto-stops, the auto-start/stop function only turns off the underlying device(s); the exposed _VTherm_ keeps the `hvac_mode` you requested (e.g. `heat` or `cool`), and its `hvac_action` switches to `off` to reflect that the equipment is currently inactive. Similarly when it auto-starts, it updates `hvac_action` to the mode it's using for the underlying climate entity.

---

This is the result with the refactoring:

Actively Cooling:

Screenshot 2026-07-03 at 10 00 55 PM
Screenshot 2026-07-03 at 10 01 05 PM

Auto-stopped:

Screenshot 2026-07-03 at 9 59 32 PM
Screenshot 2026-07-03 at 10 00 03 PM

Marking as draft since it doesn't have any unit tests yet, but I am using the fork for my system locally and it works really well! Let me know if you have any thoughts about this change and we can get it to the finish line.


jmcollin78:
Hello, any news on this PR ?
If you wait to much you have to rebase and merge. Many things have change since your PR.

PR #2016, opened 2026-07-07
fix: set_auto_regulation_mode to None left the previous regulation algo active

ianderse:
## Problem

Calling the `versatile_thermostat.set_auto_regulation_mode` service with `auto_regulation_mode: 'None'` on a VTherm `over_climate` that was configured with an active self-regulation mode (e.g. Medium) does not actually stop the regulation.

`choose_auto_regulation_mode()` sets `self._auto_regulation_mode` (so `is_regulated` correctly starts reporting `False`), but the do-nothing regulator is only installed when no algo exists yet:

```python
if not self._regulation_algo:
    # A default empty algo (which does nothing)
    self._regulation_algo = PITemperatureRegulator(self.target_temperature, 0, 0, 0, 0, 0, True)
```

Since the previously configured regulator (e.g. Medium) is still assigned, `_send_regulated_temperature()` keeps using it and keeps sending regulated setpoints to the underlying climate entities — while the entity attributes claim regulation is off.

## Real-world impact

I use the service to freeze regulation overnight because my Matter AC units relight their display LED on every command received. After calling `set_auto_regulation_mode: 'None'` (and disabling auto start/stop), the underlying unit still received regulated setpoint changes through the night (e.g. 73→74→75→76°F as the room cooled), stepping every `auto_regulation_periode_min` whenever the drift exceeded `auto_regulation_dtemp` — waking the display each time. Attributes showed `is_regulated: false` the whole time, which made this quite hard to diagnose.

## Fix

Always install the do-nothing regulator when the mode is None:

```python
if self._auto_regulation_mode == CONF_AUTO_REGULATION_NONE or not self._regulation_algo:
```

## Test

Added `test_over_climate_set_regulation_mode_none_replaces_algo`: creates a Medium-regulated over_climate VTherm, asserts the regulator produces an offset, switches to `None` via the service, and asserts the active algo now returns exactly the target temperature. The test fails on `main` and passes with this fix; the full `tests/test_auto_regulation.py` suite passes (7/7, Python 3.14).

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -416,8 +416,10 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
                     DOMAIN,
                 )
 
-        if not self._regulation_algo:
-            # A default empty algo (which does nothing)
+        if self._auto_regulation_mode == CONF_AUTO_REGULATION_NONE or not self._regulation_algo:
+            # A default empty algo (which does nothing). It must also replace any
+            # previously active algo when switching to None at runtime, else the
+            # old regulator keeps sending regulated setpoints to the underlyings.
             self._regulation_algo = PITemperatureRegulator(self.target_temperature, 0, 0, 0, 0, 0, True)
 
     def choose_auto_fan_mode(self, auto_fan_mode: str):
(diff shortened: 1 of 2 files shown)


jmcollin78 (approved):
Hello, thank you for the PR and the explanations !

jmcollin78:
I just publish a beta with your PR merged. Please give it a test a let me know. https://github.com/jmcollin78/versatile_thermostat/releases/tag/10.1.0.beta1

Thank you !

PR #2021, opened 2026-07-13
fix(over_climate): repair hvac_mode drift and space out mode/temperature commands

### LINKED ISSUE #2020: over_climate: underlying climate can get stuck OFF forever after a spontaneous drift (no periodic hvac_mode resync)
## Describe the bug

With an `over_climate` VTherm, the underlying climate entity can silently drift away from the `hvac_mode` VTherm asked for, and **VTherm never notices nor corrects it**. Once this happens, the underlying stays stuck in the wrong mode (typically OFF) indefinitely, even though VTherm's own state still shows it as ON.

This happens because `hvac_mode` is only (re)sent to the underlying when VTherm's own internal state changes (new preset, new hvac_mode request, ...). The periodic control cycle (`async_control_heating`, every `cycle_min`) only resends the regulated *temperature* to the underlying — it never re-verifies or re-sends `hvac_mode`. So if the underlying's own reported state drifts away from what VTherm requested (device echo/bounce, cloud-polled integration lag, lost command, ...), nothing will ever bring it back in sync until the user manually toggles the VTherm's preset or hvac_mode.

## Expected behavior

If the underlying climate's actual `hvac_mode` no longer matches what VTherm currently wants (`vtherm_hvac_mode`), VTherm should detect and repair this drift on its own — periodically, without requiring the user to intervene, similar to what `check_initial_state()` already does once at startup.

## Steps to reproduce / logs

Underlying climate: `climate.17592186096528_thermostat` ("Midea Thermostat", used to control a Qlima portable AC), controlled by VTherm `Clim salon VT` (`over_climate`).

Sequence: VTherm was OFF (eco preset). The preset was switched to `comfort`, which should turn the AC ON in `cool` mode.

```
2026-07-13 19:36:02.865 INFO [...thermostat_climate] Clim salon VT - ---------------------> NEW EVENT: Underlying climate climate.17592186096528_thermostatstate changed from <state ...=off; ... hvac_action=off ...> to <state ...=cool; ... hvac_action=idle ...>
2026-07-13 19:36:02.865 DEBUG [...thermostat_climate] Clim salon VT - Underlying climate climate.17592186096528_thermostat have changed. new_hvac_mode is <no change> (vs cool), ...
2026-07-13 19:36:02.865 INFO [...thermostat_climate] Clim salon VT - underlying event is received less than 10 sec after command. Forget it to avoid loop

2026-07-13 19:36:03.356 INFO [...thermostat_climate] Clim salon VT - ---------------------> NEW EVENT: Underlying climate climate.17592186096528_thermostatstate changed from <state ...=cool; ...> to <state ...=off; ... hvac_action=off ...>
2026-07-13 19:36:03.356 DEBUG [...thermostat_climate] Clim salon VT - Underlying climate climate.17592186096528_thermostat have changed. new_hvac_mode is off (vs cool), new_hvac_action=off (vs off), ...
2026-07-13 19:36:03.356 INFO [...thermostat_climate] Clim salon VT - underlying event is received less than 10 sec after command. Forget it to avoid loop

2026-07-13 19:36:03.374 INFO [...thermostat_climate] Clim salon VT - Calling ThermostatClimate._send_regulated_temperature force=True
2026-07-13 19:36:03.374 INFO [...thermostat_climate] Clim salon VT - Regulated temp have changed to 25.2. Resend it to underlyings
```

What happened, step by step:

1. VTherm correctly sends the `cool` command once (the AC briefly reports `off → cool`).
2. Less than a second later, **the AC/Midea integration flips back to `off` on its own** — this is a device/integration-side quirk, not something VTherm did.
3. VTherm's 10-second anti-loop filter correctly ignores this bounce-back (`"underlying event is received less than 10 sec after command. Forget it to avoid loop"`), since it looks like an echo of VTherm's own just-sent command.
4. **But this is where the real bug is**: because VTherm ignores this event, it never re-checks whether the underlying actually ended up in the right state. It just moves on to resending the regulated *temperature* (`_send_regulated_temperature`), never `hvac_mode` again.
5. The AC stays `off`. VTherm itself still believes it is `cool`/on (`vtherm_hvac_mode == cool`), and nothing in the periodic control cycle (every `cycle_min`, 40 min in this config) ever resends the `hvac_mode` command to correct it. The AC remains off indefinitely — confirmed by testing and waiting past a full cycle.

So the net effect: VTherm's initial `cool` command effectively failed (the device ignored/overrode it a moment later), and VTherm has no mechanism to detect that failure and retry.

## Additional context

- This is more visible on cloud-polled / less reliable climate integrations (reported here with a Midea-based integration controlling a Qlima portable AC) than on more responsive local integrations (e.g. Mitsubishi via a local integration), where the bounce-back apparently doesn't happen.
- As a workaround, the reporter built a Home Assistant automation that checks 5 minutes after VTherm turns on whether the underlying AC actually turned on, and forces it on if not.
- Debug logs (`custom_components.versatile_thermostat: debug`) for the full sequence are available if needed.

## Versions

- Versatile Thermostat: latest (main branch, includes fix for #1994)
- Underlying: Midea-based climate integration controlling a Qlima portable AC
- VTherm type: `over_climate`, `auto_regulation_mode` enabled (PI regulation)

Yoldark34:
Fixes #2020

## Problem

Some climate integrations (cloud-polled devices like **Midea Auto Cloud**) silently drift away from the hvac_mode VTherm asked for. Investigating on real hardware revealed two distinct issues:

1. **Silent drift is never repaired**: VTherm only resends hvac_mode when its own internal state changes, so an underlying that bounces back to a previous mode (or is changed externally by remote/app) stays out of sync indefinitely.
2. **Back-to-back commands cause the bounce-back in the first place**: sending `set_temperature` immediately after `set_hvac_mode` makes the Midea device accept the mode change, then revert to its previous mode ~15 seconds later — even though the mode change alone (or the same two commands manually spaced by a few seconds) is handled fine. This defeated the drift repair itself: the repair command was sent and accepted, then undone by its own temperature follow-up.

## History of the changes and why they were done like that

The fix evolved through real-device testing; each step is a separate commit:

1. **Periodic drift resync** (`6225278` → `ad8d5a2`): extract the startup sync logic (`check_initial_state`) into a reusable `resync_hvac_mode()` and call it on every control cycle, in both directions (underlying off while VTherm wants heat/cool, and underlying on while VTherm is off). It is deliberately gated by `auto_regulation_period_min` — so retries cannot hammer a device-side compressor protection — and by a grace period after VTherm's own commands, so a stale/echoed state cache is not misread as a drift.

2. **Command spacing** (`1f12f52`): live testing showed the repair command was accepted but the device bounced back to `off` 15s later. The cause was the immediate regulated-temperature push following the mode command. When a `set_hvac_mode` was just sent to an underlying (tracked by a consume-once flag, covering both the fresh turn-on cascade and the drift-repair resync), the immediate regulated-temperature send is skipped and the delayed resend already scheduled by `set_hvac_mode` (`resend_delay_sec` = 2s, originally added for lazy SonoffTRVZB devices) carries the temperature once the device has settled.

   Two other approaches were tried and rejected:
   - a blocking `asyncio.sleep(2)` inside `set_hvac_mode`: opens a reentrancy window in the control cycle (other events can trigger nested cycles during the sleep) and broke 7 tests by mixing real wall-clock sleep with the fake test clock;
   - a timestamp-based check inside `set_hvac_mode` itself: also suppressed legitimate immediate sends on the normal turn-on path.

3. **Delayed send carries the regulated temperature** (`85059d6`): the raw-target fallback meant the device got the plain setpoint at +2s and the properly regulated value only at the next regulation period (minutes later). Now the already-computed regulated temperature (including device-temp offset) replaces the raw-target resend, so the device ends up with the right setpoint right away. Also in this commit:
   - the mode-just-sent check was moved above the dtemp gate so the consume-once flag cannot go stale (a stale flag would have spuriously skipped one send on a later cycle);
   - a direct `set_temperature` now cancels any pending delayed resend, so a stale delayed value can never overwrite a newer command;
   - the raw-target fallback (used when the thermostat is not regulated) reads the target at fire time, preserving the original SonoffTRVZB freshness semantics.

4. **Observability** (`fbdc8dc`, `7748452`): debug logs for the whole lifecycle (scheduling a delayed send, replacing a pending one, firing it, cancelling it on a direct send), plus rounding of the sent value (`a4957e6`, absorbs float noise like `23.099999999999998`) and two-decimal formatting in the logs.

## Why some unrelated-looking unit tests changed

- **`test_over_climate_regulation_ac_mode`**: this test implicitly depended on the buggy behavior. The old expected PI values (`25 - 2.5`, `25 - 1`, `25 + 3`) were the output of a regulation-cycle history that included an *extra* cycle incidentally triggered by the mock underlying echoing the (now eliminated) back-to-back mode+temperature commands. Since the PI output is stateful (`accumulated_error`, `time_delta`), removing that accidental cycle changes the exact offsets. The test now drives that cycle explicitly (`async_control_heating(force=True)` after the sensor updates), stubs the delayed resend so it doesn't race the test's fake clock, and asserts the values that match the PI math for the real cycle sequence (`25 - 2.0`, `25 - 0.5`, `25 + 3.5`). The direction of every assertion is unchanged (regulated below target when too warm, above when too cold). Also, `hvac_action` right after turn-on is now deterministically `IDLE` instead of a stale `OFF` read.

- **`test_over_climate_regulation_use_device_temp`**: pre-existing flaky test, already documented in the file ("Disable this test which is not working when run in // of others. I couldn't find out why") with a disabled `@pytest.mark.skip`. The failure reproduces on the unmodified base branch at the same rate (~1 in 8 full-suite runs). Root cause: `wait_for_local_condition` was called without `hass`, so it never flushed pending HA tasks while polling, and it returns as soon as the mock's raw python attribute flips — a beat before the corresponding `state_changed` event is processed by VTherm. Deflaked by passing `hass=hass`, flushing once more after the condition holds, and enabling the already-written-but-commented-out `calculate_hvac_action()` call.

## Validated on real hardware (Midea AC via Midea Auto Cloud)

- Fresh turn-on: mode sent → 2s pause → regulated setpoint delivered → device settles into `cool`, no bounce-back
- Drift repair, ON direction (AC turned off manually while VTherm cooling): detected and re-aligned at the next regulation window, temperature delivered 2s after the mode command
- Drift repair, OFF direction (AC turned on manually while VTherm off): detected and turned off, correctly with no temperature follow-up

## Note for future work

The periodic drift resync could be relocated into `FeatureRepairIncorrectStateManager` (active only when `is_repair_incorrect_state_configured` is true) so users can opt in/out of the repair behavior explicitly. However, the **mode/temperature command spacing must stay unconditional**: it fixes the root cause (devices reverting on back-to-back commands) by itself — including for the repair manager's own re-emitted commands, which previously could spam a device without success for this very reason.

## Tests

615 passed, 2 skipped; suite runtime unchanged (~57s). Verified stable across 18+ consecutive full-suite runs.

### DIFF
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -275,6 +275,25 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
                 )
                 continue
 
+            # Some climate integrations (e.g. Midea cloud) mishandle a set_temperature
+            # call fired right on the heels of set_hvac_mode: the device bounces back
+            # to its previous mode shortly after, even though the mode change alone
+            # (without a following command) is handled fine. If a hvac_mode command was
+            # just sent to this underlying earlier in this same control cycle (turn-on
+            # cascade or drift-repair resync), skip the immediate send here and let the
+            # delayed resend already scheduled by set_hvac_mode (see resend_delay_sec)
+            # carry it once the underlying has settled. consume_mode_just_sent() resets
+            # itself, so a later, genuinely separate temperature change is unaffected.
+            if under.consume_mode_just_sent():
+                _LOGGER.debug(
+                    "%s - hvac_mode was just sent to %s (startup/resync) -> skip sending the regulated temperature %.1f for now, the delayed resend will send the target temperature in %s sec",
+                    self,
+                    under.entity_id,
+                    target_temp,
+                    resend_delay_sec,
+                )
+                continue
+
             _LOGGER.debug(
                 "%s - The device offset temp for regulation is %.2f - internal temp is %.2f. New target is %.2f",
                 self,
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -277,7 +277,7 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
             # later, genuinely separate temperature change is unaffected.
             if under.consume_mode_just_sent():
                 _LOGGER.debug(
-                    "%s - hvac_mode was just sent to %s (startup/resync) -> delay sending the regulated temperature %.1f by %s sec",
+                    "%s - hvac_mode was just sent to %s (startup/resync) -> delay sending the regulated temperature %.2f by %s sec",
                     self,
                     under.entity_id,
                     target_temp,
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -184,6 +184,21 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
             self.do_send_regulated_temp_later()
             return
 
+        # Some climate integrations can silently drift away from the hvac_mode VTherm
+        # asked for (device bounces back to a previous mode on its own, a command is
+        # lost, a compressor minimum-off-time protection silently ignores the command,
+        # ...) since VTherm otherwise only resends hvac_mode when its own internal
+        # state changes. Realign them here: the auto-regulation period gate above
+        # already ensures we don't get here more often than that period allows (or
+        # that force=True was explicitly requested), so retrying the repair can't
+        # re-trigger a device-side protection before it has had a chance to let the
+        # device actually start. Still skip it for the few seconds right after
+        # VTherm itself just sent a hvac_mode command, since the underlying's cached
+        # state may not have caught up yet in that case.
+        if not self._last_change_time_from_vtherm or (self.now - self._last_change_time_from_vtherm).total_seconds() >= resend_delay_sec:
+            for under in self._underlyings:
+                await under.resync_hvac_mode()
+
         self._regulation_algo.set_target_temp(self.target_temperature)
 
         if not self._regulated_target_temp:
@@ -836,20 +851,9 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
 
         # Continue the normal async_control_heating
 
-        # Some climate integrations can silently drift away from the hvac_mode VTherm
-        # asked for (device bounces back to a previous mode on its own, a command is
-        # lost, a compressor minimum-off-time protection silently ignores the command,
-        # ...) since VTherm otherwise only resends hvac_mode when its own internal
-        # state changes. Realign them, but only once the auto-regulation period has
-        # elapsed (same gate as _send_regulated_temperature), not on every control
-        # cycle: retrying more often than that risks re-triggering a device-side
-        # protection before it has had a chance to let the device actually start.
-        if not self._last_change_time_from_vtherm or (self.now - self._last_change_time_from_vtherm).total_seconds() >= resend_delay_sec:
-            if self.check_auto_regulation_period_min(self.now):
-                for under in self._underlyings:
-                    await under.resync_hvac_mode()
-
-        # Send the regulated temperature to the underlyings
+        # Send the regulated temperature to the underlyings. The hvac_mode drift
+        # repair (see resync_hvac_mode) is also triggered from within this call,
+        # gated by the same auto-regulation period check.
         await self._send_regulated_temperature(force=force)
 
         if self._auto_fan_mode and self._auto_fan_mode != CONF_AUTO_FAN_NONE:
--- a/custom_components/versatile_thermostat/thermostat_climate.py
+++ b/custom_components/versatile_thermostat/thermostat_climate.py
@@ -265,6 +265,31 @@ class ThermostatOverClimate(BaseThermostat[UnderlyingClimate]):
             # will not have any effect on the device. This avoid to send too many temperature changes to the underlying.
             dtemp = target_temp - (under.last_sent_temperature if under.last_sent_temperature else 0)
 
+            # Some climate integrations (e.g. Midea cloud) mishandle a set_temperature
+            # call fired right on the heels of set_hvac_mode: the device bounces back
+            # to its previous mode shortly after, even though the mode change alone
+            # (without a following command) is handled fine. If a hvac_mode command was
+            # just sent to this underlying earlier in this same control cycle (turn-on
+            # cascade or drift-repair resync), don't send the regulated temperature
+            # immediately: re-schedule it resend_delay_sec later (replacing the
+            # raw-target resend set_hvac_mode scheduled), once the underlying has
+            # settled on the new mode. consume_mode_just_sent() resets itself, so a
+            # later, genuinely separate temperature change is unaffected.
+            if under.consume_mode_just_sent():
+                _LOGGER.debug(
+                    "%s - hvac_mode was just sent to %s (startup/resync) -> delay sending the regulated temperature %.1f by %s sec",
+                    self,
+                    under.entity_id,
+                    target_temp,
+                    resend_delay_sec,
+                )
+                under.set_temperature_later(
+                    target_temp,
+                    self._attr_max_temp,
+                    self._attr_min_temp,
+                )
+                continue
+
             if not force and abs(dtemp) < (self._auto_regulation_dtemp or 0):
                 _LOGGER.info(
                     "%s - dtemp (%.1f) is < %.1f -> forget the regulation send for %s",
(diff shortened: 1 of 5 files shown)


Yoldark34:
I've deployed the fix on my HA VT and testing it.

Yoldark34:
It was working but waiting for a new cycle (40 min at home) was too long in the hot summer. 
I´ve made it check after every 'async_control_heating' to make the repair quicker.

Yoldark34 (commented):
I checked all the modifications and i looks fine for me but i need someone who knows the project better to check it and or use the code to find another way to fix the bug.

Yoldark34:
It fixed my midea (Qlima) AC but broke my mitsubishi, i need more time to find a better solution

Yoldark34:
Update: I initially reverted this fix because I hit a false positive — turned out the actual root cause was on my side, not this code. I had **two separate VTherm `over_climate` entities both controlling the same physical AC unit** (one nominally for the bedroom, one for the desk room, sharing the same underlying `climate` entity). Since VTherm's control model assumes a single owner per underlying, the two VTherms ended up fighting each other over the AC's `hvac_mode` — each one saw the other's commands as unexplained drift and re-asserted its own, which looked like the exact bounce/drift symptom this PR targets but was actually a self-inflicted conflict.

My workaround: instead of running two VTherms against the same AC, I now use a single VTherm plus a **temperature proxy sensor** (template sensor) that I repoint to whichever room's temperature I want the AC to follow (bedroom or desk), so there's always exactly one controller for that unit.

With that hack removed, I re-tested this fix standalone against my real Midea AC and confirmed it correctly detects and repairs genuine hvac_mode drift (the device-side bounce/lag case this PR is actually meant to fix). Good to merge from my side.

I also bumped the repair log from `INFO` to `WARNING` in the latest commit — this fires when VTherm's belief about the underlying's state and its real state have diverged and had to be forcibly corrected, which isn't a routine/expected event, so it shouldn't be logged at INFO. The message now states explicitly what hvac_mode VTherm was requesting, what state the underlying was actually found in, and the command sent to realign it.

Yoldark34:
Follow-up commit (0bade80): in live testing (Midea AC), retrying `resync_hvac_mode()` on every control cycle turned out to be too aggressive — cycles can be as frequent as every ~2 minutes (temperature sensor updates), and some underlyings have their own protection (e.g. a compressor minimum-off-time) that silently refuses the `hvac_mode` command and bounces back to `off` within a second. Retrying that often kept re-triggering the device's own protection before it ever got a chance to actually start, producing an endless repair loop (and a beep on each attempt) that never let the AC turn on.

The repair now still runs every control cycle, but is gated behind `check_auto_regulation_period_min()` — the same period check already used for regulated temperature sends (typically 3–5 min, configurable via `auto_regulation_period_min`). This spaces out repair attempts enough to let the underlying's own start sequence complete before VTherm tries again.

Yoldark34:
I'm happy with the current fix. I need to time to field test on my HA with both my AC and check if it solves all the problems

Yoldark34:
Still finding stuff to fix...

Yoldark34:
Many of the problems seemed to be because of Midea Auto Cloud reporting the AC as off or not turning itself on correctly. Nevertheless the improvement done in this PR is mitigating all the Cloud based AC and should still fix the drift or incorrect state. 

I now use the local way to command the AC controlled by vterm and looking for the difference between the local and cloud to validate that the Midea Auto Cloud is flaky like the official app is.

* Again we can move the drift fix in the repair state code
* Maybe adding a numeric input value in underlings to ask for commands tootling (1 sec by default) to avoid spamming the unit ? It may be the better solution than the fix for lazy sonoff done in Fixes #1913 https://github.com/jmcollin78/versatile_thermostat/commit/6414480db9b84f7387ed5657141c8763ffeff5cf

Yoldark34:
I changed to Midea Smart AC and it work better with it (local lan control vs cloud control). Still sometimes the AC become unavailable (neither the official app or msmart-ng can query it) i don't know why and how and it is not related to vterm. 

Nevertheless the fix allowed Vterm to fix the state and the drift of the AC when i plugged off and on the Qlima AC and it became available again. 

I need to add a smart switch to automate that fix myself (for my dumb Qlima AC) as it is not related to Vterm but it still shows that the fix is working and is useful.

jmcollin78:
Hello @Yoldark34,

Thank you for this PR. I will not be able to accept it like that. There is too much change that are specific to a particular device. There is a high risk of breaking other stuff with your changes.
But I have a solution and you don't have done this for nothing. There is a mecanism known as plugin that is made to specialize the behavior. This is formely another integration which is linked to the Versatile Integration.

You have technical documentation in documentation/tech-docs directory and you can inspire from existing plugin like https://github.com/jmcollin78/vtherm_api and https://github.com/jmcollin78/vtherm_pellet_stove.

If you give this examples to you Claude code, you will be able to have a new plugin for others who uses Midea AC devices.

Of course I will be able to help if you need more explanations.

jmcollin78:
I close but we can continue to discuss if needed.

Yoldark34:
Hi @jmcollin78, Thanks for the feedback, i'm glad it serve you still. Could you point me on the changes you are proposing ? I'm also very aware of how big of a change it is and why it can't be accepted. No worry.

I will check the resources you gave me. Thanks a lot.

PR #2039, opened 2026-08-11
fix(auto-tpi): align base service signature with schema

### LINKED ISSUE #2038: set_auto_tpi_mode rejects its own declared optional parameters (reinitialise, allow_kint_boost_on_stagnation, allow_kext_compensation_on_overshoot) — 500 on every call
**Integration version:** 10.0.2 (HACS, installed via `custom_components/versatile_thermostat`)
**Home Assistant:** config-entry based VTherm setup (Central configuration entry + 15 per-zone `over_climate`/`over_switch` entries)

## Description
Calling `versatile_thermostat.set_auto_tpi_mode` fails with a 500 Internal Server Error on every VTherm zone, regardless of which optional parameters are included in the service call.

## Steps to reproduce
Call the service with any of these payloads (all fail identically):
- `{"auto_tpi_mode": true, "reinitialise": true, "allow_kint_boost_on_stagnation": false, "allow_kext_compensation_on_overshoot": false}`
- `{"auto_tpi_mode": true, "allow_kint_boost_on_stagnation": false, "allow_kext_compensation_on_overshoot": false}` (omitting reinitialise)
- `{"auto_tpi_mode": true}` (omitting all three optional fields)

## Actual behavior
500 error every time. home-assistant.log shows, in turn, for whichever field was the first optional one in play:

```
TypeError: BaseThermostat.service_set_auto_tpi_mode() got an unexpected keyword argument 'reinitialise'
  File ".../versatile_thermostat/prop_handler_tpi.py", line 500, in service_set_auto_tpi_mode
    await self._algo_handler.service_set_auto_tpi_mode(
  File ".../versatile_thermostat/thermostat_prop.py", line 291, in service_set_auto_tpi_mode
```

The same TypeError then fires for `allow_kint_boost_on_stagnation`, then `allow_kext_compensation_on_overshoot`, as each is removed from the call in turn. `BaseThermostat.service_set_auto_tpi_mode()`'s actual implementation does not appear to accept any of the three optional keyword arguments declared in `services.yaml` for this service. Because Home Assistant's own service-call layer fills in each field's declared schema default when the caller omits it, there is no way to avoid sending them — the service cannot currently be called successfully with any parameter combination.

## Expected behavior
The service should accept the optional parameters its own schema declares (`reinitialise`, `allow_kint_boost_on_stagnation`, `allow_kext_compensation_on_overshoot`), or the schema should be corrected to match what the implementation actually accepts.

## Impact
`set_auto_tpi_mode` is fully non-functional as shipped in 10.0.2 — blocks turning on Auto-TPI learning entirely.

KipK:
This fix missing attributes in basethermostat signature for set_autotpi_mode service.

fix #2038

### DIFF
--- a/custom_components/versatile_thermostat/base_thermostat.py
+++ b/custom_components/versatile_thermostat/base_thermostat.py
@@ -2157,7 +2157,13 @@ class BaseThermostat(ClimateEntity, RestoreEntity, Generic[T]):
             "This thermostat does not use TPI algorithm."
         )
 
-    async def service_set_auto_tpi_mode(self, auto_tpi_mode: bool):
+    async def service_set_auto_tpi_mode(
+        self,
+        auto_tpi_mode: bool,
+        reinitialise: bool = True,
+        allow_kint_boost_on_stagnation: bool = False,
+        allow_kext_compensation_on_overshoot: bool = False,
+    ):
         """Stub method for Auto TPI mode service on non-TPI thermostats.
 
         This service is only available for switch/valve type thermostats that use TPI algorithm.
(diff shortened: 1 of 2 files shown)


jmcollin78 (approved):
Thank you

PR #2041, opened 2026-08-14
fix: restore broken star history chart

Dessalines39394:
The star history chart in the README is currently broken because of GitHub stargazer API restrictions. This PR updates the chart links to a working alternative so the chart renders again for all users.

Updated README.md, README-cs.md, README-de.md, README-fr.md, and README-pl.md.


jmcollin78 (approved):
Thank you for this !

PR #2050, opened 2026-08-21
Fix delayed valve updates with proportional algorithm plugins

KipK:
## Summary

1) Valve schedulers received updated commands from proportional algorithm plugins, but the underlying valve reread the thermostat's stale `valve_open_percent` value. As a result, commands could remain unapplied until the next temperature update.

This change centralizes valve command validation in the over-valve thermostat and makes the scheduler use the retained value for both the physical command and the realized-power trace.

Existing regulation period, percentage threshold, rounding, and clamping protections remain unchanged.

2) This PR also prevents valve commands from being quantized a second time at automatic cycle restarts. Valve mode now keeps the integer percentage retained by the thermostat instead of recalculating it from cycle seconds, avoiding transient changes such as 3 → 2 → 3 and unnecessary TRV writes. PWM behavior for switches remains unchanged.

12:10 PM

## Tests

Added regression coverage for initial and mid-cycle valve command updates.

### DIFF
--- a/custom_components/versatile_thermostat/cycle_scheduler.py
+++ b/custom_components/versatile_thermostat/cycle_scheduler.py
@@ -260,7 +260,15 @@ class CycleScheduler:
             self.min_activation_delay,
             self.min_deactivation_delay,
         )
-        realized_on_percent = on_time_sec / self._cycle_duration_sec if self._cycle_duration_sec > 0 else 0.0
+        # A valve applies the retained position directly. Converting it back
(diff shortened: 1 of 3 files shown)


jmcollin78 (approved):
Thank you !

jmcollin78:
Les tests sont KO mais j'ai cru reconnaitre mes erreurs suite au passage à 2026.08. Je merge du coup.

PR #2057, opened 2026-09-07
Fix regulated temperature resend after climate startup

bontiv:
## Summary

- resend the last effective underlying setpoint after enabling heat or cool mode
- fall back to the regulated thermostat target, then the normal target, when no previous setpoint exists
- avoid issuing a delayed temperature command when no target is available
- add regression coverage for all fallback paths

## Problem

The delayed resend used for slow climate devices always sent the thermostat's raw target temperature. With auto-regulation enabled, this overwrote the regulated setpoint two seconds after startup and could leave the underlying climate at the wrong temperature until the next regulation period.

## Testing

- Python compilation succeeded for the modified source and test files
- `git diff --check upstream/main...HEAD` passed
- The focused pytest suite could not be run locally because the project virtual environment does not contain pytest


jmcollin78 (approved):
Thank you @bontiv  !

PR #2061, opened 2026-09-13
Fix misleading "Stopped attempting repairs" log in the repair incorrect state manager

matiasmacera:
## Problem

With "repair incorrect state" enabled, when an underlying entity keeps diverging from the state VTherm wants (for example a switch or climate that ignores or reverts the command), the log shows, once per control cycle:

```
ERROR ... <VTherm name> - RepairIncorrectStateManager: maximum repair attempts (5) reached. Stopped attempting repairs to avoid infinite loop.
```

The message says repairs were stopped, but they are not. Repairs resume a few cycles later, so a user reading the log gets the wrong picture of what VTherm is doing.

Reproduction (generic): configure a VTherm over a switch with repair enabled, make the switch revert to its previous state after every command, and let a few control cycles run. You get 5 repairs, then the "Stopped attempting repairs" error on the following cycles, then repairs again, in a loop.

## Cause

`custom_components/versatile_thermostat/feature_repair_incorrect_state_manager.py`, `check_and_repair()` (lines 108-123 on `main`):

- after `REPAIR_MAX_ATTEMPTS` consecutive repairs, each cycle increments the counter and returns without checking the underlyings;
- when the counter reaches `2 * REPAIR_MAX_ATTEMPTS`, it is reset to 0 and the repair runs again in the same call.

So the real behavior is a pause of `REPAIR_MAX_ATTEMPTS - 1` cycles (4 with the current constant) followed by a retry. The error was also logged on the cycle where repairs resume, right before the info message about the reset.

## Fix

Log messages and comments only. Control flow and counter handling are unchanged.

- The error is now logged only on the skipped cycles, and says what happens: `maximum consecutive repair attempts (5) reached. Repairs are paused (skipped cycle 1/4) and will be retried afterwards.`
- The resume cycle logs at info level: `repair pause is over, resetting the counter and retrying repairs.`
- The class docstring and inline comment describe the pause/retry instead of a cap.

## How it was tested

- Added `tests/test_repair_incorrect_state_manager.py`: a mocked VTherm with one underlying that never converges. It checks 5 repairs, then 4 skipped cycles (no underlying check, one error each, containing the pause progress and not "Stopped attempting"), then a retry with no error and the resume info message.
- The new test fails on `main` (the message contains "Stopped attempting repairs") and passes with this change.
- Full suite, as in the `Run Tests` workflow (`pytest -qq --timeout=9 -n auto ... tests`, Python 3.14): 657 passed, 2 skipped.
- `black` (line length 180 from `pyproject.toml`) reports no changes in the lines touched by this PR. The file already had unrelated black differences, which are left alone to keep the diff small.

## Possible follow-up (not in this PR)

- **Per-underlying retry counter.** The counter is per VTherm, shared by all its underlyings. While repairs are paused, the early `return` skips the check for every underlying, so one entity that keeps diverging also delays the repair of other underlyings that drifted during the same period. Tracking consecutive repairs per underlying would pause only the entity that does not converge.
- **Time-based backoff with a total cap.** Instead of a fixed pause counted in cycles, the delay between repair attempts for an entity could grow (for example doubling), with a maximum number of attempts or total duration after which VTherm stops retrying and raises a persistent, visible signal (repair issue or attribute) until the entity behaves again or the VTherm is reconfigured.

Happy to work on either if you think it fits.


jmcollin78 (approved):
Thank you !

